PR #17, opened 2023-12-05
Removed `push_github` step from bitbucket pipeline.

miladehghani:
Removed `push_github` step from bitbucket pipeline.


kev2480:
@miladehghani Apologies I meant for you to do this in our Bitbucket Repo 👌  You're right, this bitbucket pipelines file can be removed instead for this PR 🎉

kev2480 (changes requested):
See comment i added (remove this file and then remove push_github from our bitbucket repo)

miladehghani:
@kev2480 that makes more sense now, i've updated this pr to remove bitbucket-pipelines.yml file instead.

kev2480 (approved):

PR #53, opened 2024-02-12
fix: TM-418 add tree species

LimberHope:
[task link](https://gfw.atlassian.net/browse/TM-418)

### DIFF
--- a/src/components/elements/Inputs/TreeSpeciesInput/RHFTreeSpeciesInput.tsx
+++ b/src/components/elements/Inputs/TreeSpeciesInput/RHFTreeSpeciesInput.tsx
@@ -30,7 +30,7 @@ const RHFTreeSpeciesInput = (props: PropsWithChildren<RHFTreeSpeciesInputProps>)
 
   const { mutate: createTreeSpecies } = usePostV2TreeSpecies({
     onSuccess(data) {
-      const _tmp = [...value];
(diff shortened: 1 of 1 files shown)


src/components/elements/Inputs/TreeSpeciesInput/RHFTreeSpeciesInput.tsx
```diff
@@ -30,7 +30,7 @@ const RHFTreeSpeciesInput = (props: PropsWithChildren<RHFTreeSpeciesInputProps>)
 
   const { mutate: createTreeSpecies } = usePostV2TreeSpecies({
     onSuccess(data) {
-      const _tmp = [...value];
+      const _tmp = [...(value || "")];
```
roguenet, line 33: if we're spreading a value here, a better default might be [], since `value` is presumably an iterable type. Also, while we're here, there isn't much reason to do a `push` right after a spread. Also, _tmp doesn't get used for anything else so assigning it to a variable isn't super useful. I'd do these three lines like this instead:
```
onChange([...(value ?? []), data.data])
```

cesarLima1: Following your suggestion, the changes were made to the corresponding lines

cesarLima1 (approved):

PR #75, opened 2024-02-29
[TM-570] When there is an update request awaiting approval, add an icon to the report tab.

roguenet:
https://gfw.atlassian.net/browse/TM-570


pachonjcl (approved):
LGTM

PR #84, opened 2024-03-05
[TM-701] Allow report titles to wrap.

roguenet:
https://gfw.atlassian.net/browse/TM-701


roguenet:
I hadn't realized this had been previously fixed because the epic branch was behind `staging`. I like simply removing the class more than changing it to `whitespace-normal`, so I'm leaving this PR open, but note that now that I've updated the epic branch, I've seen that this PR is a bit superfluous.

PR #114, opened 2024-04-11
Change login view

LimberHope:
login to terramatch


roguenet:
According to Vercel, this branch is failing to compile:
```
./src/components/componentsToLogin/Input/InputWrapper.tsx:28:61
Type error: 'props.variant' is possibly 'undefined'.
  26 | const InputWrapper: React.FC<InputWrapperProps> = props => {
  27 |   return (
> 28 |     <div className={`${props.containerClassName} relative ${props.variant.constent} flex flex-col gap-2`}>
     |                                                             ^
  29 |       <InputLabel
  30 |         htmlFor={props.inputId}
  31 |         required={props.required}
  ```

PR #169, opened 2024-05-15
Feat/ Add commentary box and commentary

diego-morales-flores-1996:
Description:

Added new components for comments functionality, including Commentary and CommentaryBox. Additionally, included a file with mocked data to facilitate stories.


src/components/elements/Commentary/Commentary.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const statusStyle = {
+  Submitted: { container: "bg-primary-200", textColor: "text-primary" },
+  Draft: { container: "bg-pinkCustom-200", textColor: "text-pinkCustom" }
+};
+
+const Comentary = (props: ComentaryProps) => {
+  const { name, lastName, date, comentary, files = [], status } = props;
+  const statusKey = status as keyof typeof statusStyle;
```
roguenet, line 24: This type should be in the CommentaryProps interface instead of simply `string`, then you don't need two variables that hold the same value, and consumers of the component know what the valid values are.

src/components/elements/CommentaryBox/CommentaryBox.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <Icon name={IconNames.PAPER_CLIP} className="h-4 w-4" />
+        </label>
+        <When condition={buttonSendOnBox}>
+          <Button variant="text" iconProps={{ name: IconNames.SEND, className: "h-4 w-4 text-darkCustom" }} />
+        </When>
+      </div>
+      <When condition={!buttonSendOnBox}>
+        <Button className="self-end" iconProps={{ name: IconNames.SEND, className: "h-4 w-4" }}>
+          <Text variant="text-12-bold" className="text-white">
+            SEND
```
roguenet, line 48: This should get wrapped in `t()` from `useT`

src/components/elements/Commentary/Commentary.tsx
diego-morales-flores-1996: done

roguenet (approved):

PR #196, opened 2024-05-22
fixes issues, data types and call mutations

LimberHope:
fixes issues, data types and call mutations

### DIFF
--- a/src/admin/components/ResourceTabs/AuditLogTab/components/SiteAuditLogPolygonStatus.tsx
+++ b/src/admin/components/ResourceTabs/AuditLogTab/components/SiteAuditLogPolygonStatus.tsx
@@ -1,4 +1,4 @@
-import { FC, Fragment, useEffect, useState } from "react";
+import { FC, Fragment, useMemo } from "react";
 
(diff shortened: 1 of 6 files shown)


src/admin/components/ResourceTabs/AuditLogTab/components/SiteAuditLogPolygonStatus.tsx
```diff
@@ -47,13 +47,21 @@ const SiteAuditLogPolygonStatus: FC<SiteAuditLogPolygonStatusProps> = ({
...
     return new Intl.DateTimeFormat("en-US", options).format(new Date(date));
   };
   const mutateComment = fetchPostV2AuditStatus;
+  useEffect(() => {
+    setPolygonData({
+      uuid: record?.uuid,
+      status: record?.meta,
+      title: record?.title
+    });
+  }, [record]);
```
pachonjcl, line 64:
```
const polygonData = useMemo(() => { return {uuid ...} }, [record])

pachonjcl (changes requested):

pachonjcl (approved):

PR #197, opened 2024-05-22
Tm 755 add total fields to site reports ppc in admin view

LimberHope:
[task](https://gfw.atlassian.net/browse/TM-755)

### DIFF
--- a/src/admin/components/ResourceTabs/InformationTab/index.tsx
+++ b/src/admin/components/ResourceTabs/InformationTab/index.tsx
@@ -133,7 +133,7 @@ const InformationTab: FC<IProps> = props => {
                     >
                       <Card sx={{ padding: 3 }}>
                         <Typography variant="h6" component="h3" className="capitalize">
-                          Total Tree Species
+                          Total Trees Planted
                         </Typography>
                         {record?.total_trees_planted_count}
                       </Card>
--- a/src/admin/components/ResourceTabs/InformationTab/index.tsx
+++ b/src/admin/components/ResourceTabs/InformationTab/index.tsx
@@ -8,7 +8,11 @@ import { MonitoringPartnersTable } from "@/admin/components/ResourceTabs/Informa
 import SeedingsTable from "@/admin/components/Tables/SeedingsTable";
 import { setDefaultConditionalFieldsAnswers } from "@/admin/utils/forms";
 import List from "@/components/extensive/List/List";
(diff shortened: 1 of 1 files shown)


roguenet (approved):

src/admin/components/ResourceTabs/InformationTab/index.tsx
```diff
@@ -114,10 +126,30 @@ const InformationTab: FC<IProps> = props => {
                   </Card>
 
                   <When condition={record}>
+                    <When
+                      condition={
+                        record.framework_key === "ppc" && (props.type === "sites" || props.type === "site-reports")
```
pachonjcl, line 131: you could use props.type.startsWith('site')

src/admin/components/ResourceTabs/InformationTab/index.tsx
```diff
@@ -53,6 +57,14 @@ const InformationTab: FC<IProps> = props => {
...
   });
 
+  const { data: seedlings } = useGetV2SeedingsENTITYUUID({
+    pathParams: {
+      uuid: record?.uuid,
+      entity: resource.replace("Report", "-report")
+    }
+  });
+
+  const totalSeedlings = seedlings?.data?.reduce((acc, curr) => acc + (curr?.amount ?? 0), 0);
```
pachonjcl, line 67: would be better to have a `useMemo` here to avoid recompute on every render

src/admin/components/ResourceTabs/InformationTab/index.tsx
pachonjcl: also for the `record.framework_key === "ppc"` please use `useFramework` custom hook.

src/admin/components/ResourceTabs/InformationTab/index.tsx
```diff
@@ -114,10 +126,30 @@ const InformationTab: FC<IProps> = props => {
...
+                          Total Trees Planted
+                        </Typography>
+                        {record?.total_trees_planted_count}
+                      </Card>
+                    </When>
                     <TreeSpeciesTable uuid={record.uuid} entity={resource} />
                   </When>
 
                   <When condition={props.type === "sites" || props.type === "site-reports"}>
+                    <When condition={record.framework_key === "ppc"}>
```
pachonjcl, line 145: same as above about use `useFramework`

pachonjcl (changes requested):

PR #275, opened 2024-06-24
[TM-997] Fix: Ensure "record.uuid" is safely accessed in getSiteBbox function

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-997) TM-997
[Task Link](https://gfw.atlassian.net/browse/TM-988) TM-988


roguenet (commented):
If the record doesn't exist, won't this API response not be useful? It seems like this should also have an `enabled` flag
```
{
  pathParams: { record?.uuid },
},
{
  enabled: record?.uuid != null
}
```

roguenet (approved):

PR #290, opened 2024-06-27
hotfix: null coalescing on restoration_strategy project detail

pachonjcl:
A null value that I found while testing projects.


roguenet (approved):
FWIW, I wouldn't consider this a hotfix. Stuff going into the release branch is just stuff going into the release. A hotfix would be a release directly to production to fix an urgent bug.

PR #297, opened 2024-06-27
[TM-1035] hiding unused elements

diego-morales-flores-1996:
[Task Link](https://gfw.atlassian.net/browse/TM-1035)


roguenet (commented):
Please remove the code instead of commenting it out. Even if it's temporary, the git history is there to pull things out that need to be resurrected and commented out code is cruft we'd prefer to avoid.

roguenet (approved):

PR #308, opened 2024-06-28
[TM-1045] assign draft on create

egrojMonroy:
https://gfw.atlassian.net/browse/TM-1045


roguenet (approved):
Code-wise, this seems fine; but why is the FE even assigning statuses? That should be a BE concern.

PR #329, opened 2024-07-02
Feat/tm 1069 add loader

dottyy:
[Task link](https://gfw.atlassian.net/jira/software/c/projects/TM/boards/40?selectedIssue=TM-1069)


src/context/loaderAdmin.provider.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const hideLoader = () => setLoading(false);
+
+  const value = useMemo(
+    () => ({
+      loading,
+      showLoader,
+      hideLoader
+    }),
+    [loading]
+  );
```
roguenet, line 28: :art: I'm a little surprised the linter isn't complaining about this. It's correct, but usually the linter complains if there are any values missing from the dependency array. A more idiomatic way to structure this might be:
```
const showLoader = useCallback(() => setLoading(true));
const hideLoader = useCallback(() => setLoading(false));
const value = { loading, showLoader, hideLoader }
```
or 
```
const value = useMemo(
  () => ({
    loading,
    showLoader: () => setLoading(true),
    hideLoader: () => setLoading(false),
  }),
  [loading]
);
```

roguenet (commented):
Since the build isn't complaining, I'm going to go ahead and approve. Feel free to leave as-is and merge if you'd like.

roguenet (approved):

PR #336, opened 2024-07-03
[TM-1081] change status text Draft -> Started

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1081)

### DIFF
--- a/src/admin/components/Fields/CustomChipField.tsx
+++ b/src/admin/components/Fields/CustomChipField.tsx
@@ -24,7 +24,7 @@ const CustomChipField: React.FC<Omit<FunctionFieldProps, "render">> = props => (
           STATUS_CLASSNAME_MAP[record.readable_status] ?? "bg-grey-200 text-grey-500"
         )}
       >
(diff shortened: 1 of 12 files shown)


src/admin/components/Fields/CustomChipField.tsx
```diff
@@ -24,7 +24,7 @@ const CustomChipField: React.FC<Omit<FunctionFieldProps, "render">> = props => (
           STATUS_CLASSNAME_MAP[record.readable_status] ?? "bg-grey-200 text-grey-500"
         )}
       >
-        {record.readable_status == "Started" ? "Draft" : record.readable_status}
+        {record.readable_status == "Unknown" ? "Started" : record.readable_status}
```
roguenet, line 27: Hmm, we shouldn't get an "Unknown" status from the server. That indicates something that needs fixing in the BE.

roguenet (approved):
I'll go ahead an approve, but please investigate why the server is sending "Unknown". That's likely a missing mapping in the BE model `$statuses` array.

PR #354, opened 2024-07-09
[TM-1091] Fix/ polygon checks firefox

dottyy:
[Task link](https://gfw.atlassian.net/jira/software/c/projects/TM/boards/40?selectedIssue=TM-1091)


roguenet:
This looks good, but snapshots need to be regenerated before it can merge.

PR #364, opened 2024-07-10
[TM-1099] add restoration in progress button on site admin

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-1099)

### DIFF
--- a/src/admin/components/Dialogs/StatusChangeModal.tsx
+++ b/src/admin/components/Dialogs/StatusChangeModal.tsx
@@ -59,30 +59,43 @@ const StatusChangeModal = ({ handleClose, status, ...dialogProps }: StatusChange
   })();
 
   const dialogTitle = (() => {
-    let title = status === "approve" ? "Are you sure you want to approve this " : "Request more information for ";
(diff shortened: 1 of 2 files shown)


src/admin/components/Dialogs/StatusChangeModal.tsx
```diff
@@ -59,30 +59,43 @@ const StatusChangeModal = ({ handleClose, status, ...dialogProps }: StatusChange
...
+        name = record?.title ?? "Site Report";
         break;
       case "nurseryReport":
-        title += record?.title ?? "Nursery Report";
+        name = record?.title ?? "Nursery Report";
         break;
     }
 
-    return title;
+    return title.replace("{}", name);
```
roguenet, line 98: :art: This replace is a little weird. I'd do it instead by moving the `title` block down here and in each case of that block just do something like 
```
return `Request more information for ${name}`;
```

roguenet (dismissed):

roguenet (dismissed):

roguenet (approved):

PR #379, opened 2024-07-16
[TM-1105] Framework display refactor

roguenet:
https://gfw.atlassian.net/browse/TM-1105


src/context/framework.provider.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const { frameworksShow, frameworksHide, ...rest } = props;
+    if (!useFrameworkShowHide({ frameworksShow, frameworksHide })) return null;
+
+    return <WrappedComponent {...(rest as T & JSX.IntrinsicAttributes)} />;
+  };
+  FrameworkShowHide.displayName = `withFrameworkShow(${displayName})`;
+  return FrameworkShowHide;
+}
+
+export default FrameworkProvider;
```
roguenet, line 67: This file is the main workhorse of the new system. It provides:
* A new react context that provides the current Framework (as defined in the enum in this file) to components in the render tree
* a hook (`useFrameworkContext`) to get the current contextual framework.
* an HOC (`withFrameworkShow`) to wrap a component with the logic to show or hide that component based on the current framework.

src/context/framework.provider.tsx
```diff
@@ -0,0 +1,67 @@
...
+  PPC = "ppc",
+  TF = "terrafund",
+  TF_LANDSCAPES = "terrafund-landscapes",
+  ENTERPRISES = "enterprises",
+  HBF = "hbf",
+
+  UNDEFINED = "undefined"
+}
+
+export const ALL_TF = [Framework.TF, Framework.TF_LANDSCAPES, Framework.ENTERPRISES];
```
roguenet, line 13: Previous code in some cases defined Terrafund as "not HBF or PPC", so including the three TF frameworks in one array makes replacing code like that simple.

src/hooks/useFrameworkTitle.ts
```diff
@@ -0,0 +1,25 @@
...
+    case Framework.TF:
+    case Framework.TF_LANDSCAPES:
+    case Framework.ENTERPRISES:
+      return t("TerraFund");
+
+    case Framework.UNDEFINED:
+    default:
+      return null;
+  }
+};
```
roguenet, line 25: This hook should be used in all cases where we want to display the current framework name, instead of redefining it in lots of places.

src/context/ContextCondition.tsx
```diff
@@ -0,0 +1,10 @@
+import { ReactNode } from "react";
+
+import { withFrameworkShow } from "@/context/framework.provider";
+
+/**
+ * A simple component that displays children, but is wrapped in HOCs that add contextual conditional
+ * control. Currently only uses `withFrameworkShow`, but if this pattern is repeated in the future,
+ * further HOCs may be added.
+ */
+export const ContextCondition = withFrameworkShow(({ children }: { children: ReactNode }) => <>{children}</>);
```
roguenet, line 10: This component is very simple, but I think it will expand over time. The intent here is to provide a replacement for the If/Then and When structures from react-if when we simply want to show or hide content based on framework. There are many examples of its use in this PR.

It's primarily useful for two cases:
 * Showing or hiding multiple components in a block based on framework
 * Adding show/hide based on framework functionality to components that we didn't define ourselves (this is used heavily in the admin pages).
 
 As noted in the comment above, if we end up expanding on this pattern to provide more HOCs that control child inclusion based on some contextual information, we can add those HOCs to this component to expand its usefulness.

pachonjcl (approved):

cesarLima1:
Everything seems to be okay. Just wondering, in case we add a new framework, do we have to edit the next components:
framework.provider.tsx
useFrameworkTitle.ts
right?

roguenet:
> Everything seems to be okay. Just wondering, in case we add a new framework, do we have to edit the next components: framework.provider.tsx useFrameworkTitle.ts right?

Yes, that's right. That is one downside here: it was a little easier to be dynamic about new frameworks in the old system, but that was a bit of a lie. When we added HBF, we found that there were lots of code changes needed; it wasn't as simple as just creating the framework in the DB. So, in this case I'd much rather have explicit guardrails around framework creation than to have all these hidden corners where things will break when a new framework is added.

PR #388, opened 2024-07-23
[TM-1104] CRUD for project and pitch polygons

egrojMonroy:
https://gfw.atlassian.net/browse/TM-1104

Project view, edit project, draw, upload.


src/components/elements/Inputs/Map/RHFMap.tsx
```diff
@@ -46,6 +69,57 @@ const RHFMap = ({
...
+
+  const {
+    data: projectPolygon,
+    refetch: reloadProjectPolygonData,
+    isRefetching
+  } = useGetV2TerrafundProjectPolygon(
+    {
+      queryParams: {
+        entityType: entity?.entityName || "",
+        uuid: entity?.entityUUID || ""
```
roguenet, line 86: Please use `??` for defaults

src/components/elements/Inputs/Map/RHFMap.tsx
```diff
@@ -46,6 +69,57 @@ const RHFMap = ({
...
+    {
+      enabled: !!entity,
+      staleTime: 0,
+      cacheTime: 0
+    }
+  );
+  const setBbboxAndZoom = async () => {
+    if (projectPolygon?.project_polygon?.poly_uuid) {
+      const bbox = await fetchGetV2TerrafundPolygonBboxUuid({
+        pathParams: { uuid: projectPolygon.project_polygon?.poly_uuid || "" }
```
roguenet, line 98: `??`, although I'm wondering: do you really want to call this API if there is no poly_uuid available? If not, this might be better accomplished using `enabled: projectPolygon.project_polygon?.poly_uuid != null` on the query arguments, or something similar as in the case above.

src/components/elements/Inputs/Map/RHFMap.tsx
```diff
@@ -46,6 +69,57 @@ const RHFMap = ({
...
+    isRefetching
+  } = useGetV2TerrafundProjectPolygon(
+    {
+      queryParams: {
+        entityType: entity?.entityName || "",
+        uuid: entity?.entityUUID || ""
+      }
+    },
+    {
+      enabled: !!entity,
```
roguenet, line 90: Do you really want to be calling this API if entityName or entityUUID are null? Assuming not, I'd change this to
```
enabled: entity?.entityName != null && entity?.entityUUID != null
```
And then you could drop the default `?? ""` above because the API won't be called unless these are set.

src/components/elements/Map-mapbox/MapControls/PolygonHandler.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  const openFormModalHandlerAddPolygon = () => {
+    openModal(
+      <ModalAdd
+        title="Add Polygons"
+        descriptionInput={`Drag and drop a GeoJSON, Shapefile, or KML.`}
+        descriptionList={
+          <div className="mt-9 flex">
+            <Text variant="text-12-bold">TerraMatch upload limits:&nbsp;</Text>
+            <Text variant="text-12-light">50 MB per upload</Text>
```
roguenet, line 141: make sure all free text is translated in this component. I see these lines and `content` and `primaryButtonText` below at least needing an update.

src/components/elements/Map-mapbox/MapControls/PolygonModifier.tsx
```diff
@@ -0,0 +1,100 @@
...
+  return (
+    <ControlButtonsGroup direction="col" className="z-10 w-auto">
+      <IconButton
+        iconProps={{ name: IconNames.EDIT, width: 24, height: 24 }}
+        onClick={() => {
+          setIsEditing(true);
+          onClick?.();
+        }}
+        className="rounded-b-none rounded-t-lg p-[10px]"
+        aria-label="Edit Polygon"
```
roguenet, line 66: Make sure to translate text. I see this and the `Save` and `Cancel` text below at least that need an update.

egrojMonroy: done

src/components/elements/Map-mapbox/MapControls/PolygonHandler.tsx
egrojMonroy: its ready

src/components/elements/Inputs/Map/RHFMap.tsx
egrojMonroy: done

src/components/elements/Inputs/Map/RHFMap.tsx
egrojMonroy: The if above verifies if its null. but the default validation (??) is added anyway.

src/components/elements/Map-mapbox/MapControls/PolygonModifier.tsx
roguenet: Looks like this file still needs doing.

egrojMonroy: Sorry, forget that one. 
Already checked again.

roguenet (approved):

PR #397, opened 2024-07-26
[TM-1148] refactor modal provider to multiple modals

dottyy:
[Task link](https://gfw.atlassian.net/jira/software/c/projects/TM/boards/40?selectedIssue=TM-1148)


src/admin/components/ResourceTabs/PolygonReviewTab/components/PolygonStatus/StatusDisplay.tsx
```diff
@@ -244,17 +245,19 @@ const StatusDisplay = ({
   const onFinallyRequest = () => {
     refresh?.();
     reloadEntity?.();
-    closeModal();
+    closeModal(ModalId.STATUS_CHANGE);
+    closeModal(ModalId.CHANGE_REQUEST);
```
pachonjcl, line 249: why it has to call twice?

src/components/elements/Map-mapbox/MapControls/PolygonHandler.tsx
```diff
@@ -101,7 +102,8 @@ export const PolygonHandler = () => {
       }
     }
     hideLoader();
-    closeModal();
+    closeModal(ModalId.ADD_POLYGONS);
+    closeModal(ModalId.CONFIRM_POLYGON_TYPE);
```
pachonjcl, line 106: same question as above, why call it twice?

pachonjcl (changes requested):

src/admin/components/ResourceTabs/PolygonReviewTab/components/PolygonStatus/StatusDisplay.tsx
dottyy: The reason closeModal is called twice is because these modals are being opened and managed in different files, each with a unique ID. Specifically:

The modal with ID STATUS_CHANGE is opened and closed in certain files.
The modal with ID CHANGE_REQUEST is managed in a different set of files.
By ensuring that both modals are closed here, we prevent potential issues where one of the modals might remain open if only one ID was specified. This approach ensures that no matter which modal was opened, both are properly closed. This is a safeguard given that we have around five files calling STATUS_CHANGE and another three calling CHANGE_REQUEST.

Additionally, this does not generate any errors or unexpected behavior as the closeModal function gracefully handles the closure of modals based on their IDs.

src/components/elements/Map-mapbox/MapControls/PolygonHandler.tsx
dottyy: The reason closeModal is called twice is because these modals are being opened and managed in different files, each with a unique ID. Specifically:

The modal with ID ADD_POLYGONS is opened and closed in certain files.
The modal with ID CONFIRM_POLYGON_TYPE is managed in a different set of files.
By ensuring that both modals are closed here, we prevent potential issues where one of the modals might remain open if only one ID was specified. This approach ensures that no matter which modal was opened, both are properly closed. This is a safeguard given that we have around five files calling STATUS_CHANGE and another three calling CHANGE_REQUEST.

Additionally, this does not generate any errors or unexpected behavior as the closeModal function gracefully handles the closure of modals based on their IDs.

dottyy: done, remove CloseModal unnecessary

pachonjcl (approved):

PR #401, opened 2024-07-31
Feat/tm 1144 output description validation

egrojMonroy:
https://gfw.atlassian.net/browse/TM-1144

display correct messaging with transifex templates

### DIFF
--- a/src/components/elements/MapPolygonPanel/ChecklistInformation.tsx
+++ b/src/components/elements/MapPolygonPanel/ChecklistInformation.tsx
@@ -92,9 +92,14 @@ const ChecklistInformation = () => {
                 </Text>
                 {item.messages &&
                   item.messages.map((info: any, index: number) => (
-                    <Text key={index} variant="text-14-light">
-                      {t(info)}
-                    </Text>
+                    <div className="flex items-start gap-[6px] pl-6" key={index}>
+                      <div className="mt-[3px] flex items-start lg:mt-[4px] wide:mt-[6px]">
+                        <span className="text-[7px] text-white">&#9679;</span>
+                      </div>
+                      <Text variant="text-10-light" className="text-white ">
+                        {t(info)}
+                      </Text>
+                    </div>
                   ))}
               </>
             ))}
--- a/src/components/elements/MapPolygonPanel/ChecklistInformation.tsx
+++ b/src/components/elements/MapPolygonPanel/ChecklistInformation.tsx
@@ -82,9 +82,9 @@ const ChecklistInformation = () => {
       <div className="mt-3 grid gap-3">
         <If condition={validationStatus}>
           <Then>
-            {polygonValidationData.map((item, index) => (
-              <>
-                <Text variant="text-14-light" key={index} className="flex items-center gap-2">
+            {polygonValidationData.map(item => (
+              <div key={item.id}>
+                <Text variant="text-14-light" className="flex items-center gap-2">
                   <Icon
                     name={item.status ? IconNames.CHECK_PROGRESSBAR : IconNames.IC_ERROR_PANEL}
                     className={`h-4 w-4 lg:h-5 lg:w-5 ${item.status ? "text-green-400" : ""}`}
@@ -92,8 +92,8 @@ const ChecklistInformation = () => {
                   {t(item.label)}
                 </Text>
                 {item.extra_info &&
-                  getFormatedExtraInfo(item.extra_info, item.id).map((info: any, index: number) => (
-                    <div className="flex items-start gap-[6px] pl-6" key={index}>
+                  getFormatedExtraInfo(item.extra_info, item.id).map(info => (
+                    <div className="flex items-start gap-[6px] pl-6" key={`${info}-${item.id}`}>
                       <div className="mt-[3px] flex items-start lg:mt-[4px] wide:mt-[6px]">
                         <span className="text-[7px] text-white">&#9679;</span>
                       </div>
@@ -102,7 +102,7 @@ const ChecklistInformation = () => {
                       </Text>
                     </div>
                   ))}
-              </>
+              </div>
             ))}
           </Then>
           <Else>
--- a/src/components/elements/MapPolygonPanel/ChecklistInformation.tsx
+++ b/src/components/elements/MapPolygonPanel/ChecklistInformation.tsx
@@ -93,7 +93,7 @@ const ChecklistInformation = () => {
                 {item.messages &&
                   item.messages.map((info: any, index: number) => (
                     <Text key={index} variant="text-14-light">
-                      {info}
+                      {t(info)}
                     </Text>
                   ))}
               </>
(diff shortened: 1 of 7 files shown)


src/components/elements/MapPolygonPanel/ChecklistInformation.tsx
```diff
@@ -81,13 +83,26 @@ const ChecklistInformation = () => {
...
-                  name={item.status ? IconNames.CHECK_PROGRESSBAR : IconNames.IC_ERROR_PANEL}
-                  className={`h-4 w-4 lg:h-5 lg:w-5 ${item.status ? "text-green-400" : ""}`}
-                />
-                {t(item.label)}
-              </Text>
+              <>
+                <Text variant="text-14-light" key={index} className="flex items-center gap-2">
+                  <Icon
+                    name={item.status ? IconNames.CHECK_PROGRESSBAR : IconNames.IC_ERROR_PANEL}
+                    className={`h-4 w-4 lg:h-5 lg:w-5 ${item.status ? "text-green-400" : ""}`}
```
pachonjcl, line 90: use `classnames` library

src/components/elements/MapPolygonPanel/ChecklistInformation.tsx
```diff
@@ -81,13 +83,26 @@ const ChecklistInformation = () => {
...
+                  />
+                  {t(item.label)}
+                </Text>
+                {item.extra_info &&
+                  getFormatedExtraInfo(item.extra_info, item.id).map((info: any, index: number) => (
+                    <div className="flex items-start gap-[6px] pl-6" key={index}>
```
pachonjcl, line 96: typically is not a good practice to have the index as an id, can you use another id?

src/hooks/useMessageValidations.ts
```diff
@@ -0,0 +1,118 @@
...
+        plantend: "Plant End",
+        practice: "Restoration Practice",
+        target_sys: "Target Land Use System",
+        distr: "Tree Distribution",
+        num_trees: "Number of Trees"
+      };
+      const infoArray: ExtraInfoItem[] = JSON.parse(extraInfo);
+      return infoArray
+        .map(info => {
+          if (info.exists === false) {
```
pachonjcl, line 60: can you just use `! info.exists?`

src/hooks/useMessageValidations.ts
```diff
@@ -0,0 +1,118 @@
...
+            return t("{field} is missing.", { field: fieldsToValidate[info.field] });
+          } else if (info.exists === true && info.field === "target_sys") {
```
pachonjcl, line 62: just use `info.exists` as part of the condition.

src/hooks/useMessageValidations.ts
```diff
@@ -0,0 +1,118 @@
...
+            return t(
+              "{field}: {error} is not a valid {field} because it is not one of ['agroforest', 'natural-forest', 'mangrove', 'peatland', 'riparian-area-or-wetland', 'silvopasture', 'woodlot-or-plantation', 'urban-forest'].",
+              { field: fieldsToValidate[info.field], error: info.error }
+            );
+          } else if (info.exists === true && info.field === "distr") {
```
pachonjcl, line 67: same here

src/hooks/useMessageValidations.ts
```diff
@@ -0,0 +1,118 @@
...
+        "Project Goal: Sum of all project polygons {sum_area} is {percentage}% of total hectares to be restored {total_area_project}",
+        { sum_area, percentage, total_area_project }
+      );
+    } catch {
+      return [t("Error parsing extra info.")];
+    }
+  };
+  const getFormatedExtraInfo = (extraInfo: string | undefined, criteria_id: any) => {
+    if (criteria_id === 12) {
+      return [getProjectGoalMessage(extraInfo)];
```
pachonjcl, line 106: ok, so the problem with this is when you error, you will have a `[[element]]` right?
Either return an array or a single element, but do not mix up

pachonjcl (changes requested):

src/hooks/useMessageValidations.ts
egrojMonroy: Changed

src/hooks/useMessageValidations.ts
egrojMonroy: done

src/hooks/useMessageValidations.ts
egrojMonroy: done

src/hooks/useMessageValidations.ts
egrojMonroy: done

src/components/elements/MapPolygonPanel/ChecklistInformation.tsx
dottyy: done

egrojMonroy: changed by Dotty

src/hooks/useMessageValidations.ts
```diff
@@ -0,0 +1,120 @@
...
+            });
+          } else if (info.exists && info.field === "practice") {
+            return t(
+              "{field}: {error} is not a valid {field} because it is not one of ['tree-planting', 'direct-seeding', 'assisted-natural-regeneration'].",
+              { field: fieldsToValidate[info.field], error: info.error }
+            );
+          }
+          return null;
+        })
+        .filter((message): message is string => message !== null);
```
pachonjcl, line 85: just one more thing that I noticed.
```
if (! info.exists) {
   return t(...
}
//here we are sure that info.exists is true, no need to keep checking
switch (info.field)
   case target_sys:
       return t(...
       
```

src/hooks/useMessageValidations.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (criteria_id === 12) {
+      return getProjectGoalMessage(extraInfo);
+    } else if (criteria_id === 3) {
+      return getIntersectionMessages(extraInfo);
+    } else if (criteria_id === 14) {
+      return getDataMessage(extraInfo);
+    } else {
+      return [];
+    }
+  };
```
pachonjcl, line 116: since we are inside a hook, I think you should store this value with `useMemo`.
Anyways, my main concern is for you to check that this is not being called on every render, take a time to test that this is only called once per item, and not once every render.
once you send it for review again, I will test on my end to check that.

pachonjcl (changes requested):

src/hooks/useMessageValidations.ts
egrojMonroy: you are right, changed done

src/hooks/useMessageValidations.ts
egrojMonroy: Added the useMemo, not really shure how to test. But seems that is not calling on every render now.

pachonjcl (approved):

PR #410, opened 2024-08-05
[TM-1111] add warning messages for uploads

cesarLima1:
https://gfw.atlassian.net/browse/TM-1111


src/components/extensive/Modal/ModalAdd.tsx
```diff
@@ -67,6 +71,54 @@ const ModalAdd: FC<ModalAddProps> = ({
...
+      size: file.size,
+      url: "https://google.com",
+      created_at: "now",
+      uuid: file.name,
+      is_public: true,
+      rawFile: file
+    });
+
+    const acceptedFileTypes = acceptedTYpes
+      ? acceptedTYpes
```
roguenet, line 90: Please fix the spelling (`acceptedTypes`). Also I'm curious about the join + split... isn't that a NOOP? Or are you doing it to effectively cast the values in the starting array to string? If so, I'd do this in the examples below instead of doing multiple array operations.
```
type => `${type}`.trim() 
```

🎨 I'd skip the ternary here:
```
const acceptedFileTypes = acceptedTypes?.map(type => type.trim()) ?? []
```
or 
```
const acceptedFileTypes = (acceptedTypes ?? []).map(type => type.trim())
```

src/components/extensive/Modal/ModalAdd.tsx
```diff
@@ -67,6 +71,54 @@ const ModalAdd: FC<ModalAddProps> = ({
...
+          .map(type => type.trim())
+      : [];
+
+    const filteredFiles = files.filter(file => {
+      if (
+        acceptedFileTypes.length > 0 &&
+        !acceptedFileTypes.some(type => file.type === type || file.name.endsWith(type))
+      ) {
+        setErrorMessage?.(`Unsupported file type. Please upload SHP (zipped), KML, or GeoJSON files only`);
+        return false;
```
roguenet, line 102: It seems like if we somehow end up with 0 accepted file types, that should result in an error. It's also weird to me that the error message has the set of accepted types hardcoded when it's being passed in as a prop on the component.

src/components/extensive/Modal/ModalAdd.tsx
cesarLima1: spelling changed in all related files and accepted the suggestion.

src/components/extensive/Modal/ModalAdd.tsx
cesarLima1: When no accepted type is specified, it can accept any type. and the error message now shows the accepted type dinamically.

roguenet (approved):

dottyy (dismissed):

roguenet (approved):

PR #429, opened 2024-08-10
[TM-1103] Update user roles and permissions

roguenet:
https://gfw.atlassian.net/browse/TM-1103

[BE PR](https://github.com/wri/wri-terramatch-api/pull/398)

Note that this FE PR won't pass build on Vercel until the BE PR is merged, due to API updates.

### DIFF
--- a/src/pages/auth/signup/index.page.tsx
+++ b/src/pages/auth/signup/index.page.tsx
@@ -98,7 +98,7 @@ const SignUpPage = ({
     });
   };
 
-  return <SignUpForm form={form} handleSave={handleSave} loading={isLoading} role={role as string} />;
+  return <SignUpForm form={form} handleSave={handleSave} loading={isLoading} role={role} />;
 };
 
 export default SignUpPage;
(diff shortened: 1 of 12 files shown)


src/pages/auth/signup/index.page.tsx
```diff
@@ -91,14 +91,14 @@ const SignUpPage = ({
...
-        primary_role: "project-developer",
+        role: "project-developer",
         country: selectedTitleOption == "Select Country" ? (selectedOption as any) : null,
         program: selectedTitleOption == "Select Framework" ? (selectedOption?.toLowerCase() as any) : null
       }
     });
   };
 
-  return <SignUpForm form={form} handleSave={handleSave} loading={isLoading} primary_role={primary_role as string} />;
+  return <SignUpForm form={form} handleSave={handleSave} loading={isLoading} role={role as string} />;
```
cesarLima1, line 101: Is it possible to remove the `as string` from the `role` property? Since it's already defined as a string, type assertion may be unnecessary.

roguenet: I'll give it a try; I figured if it was needed before it might still be, but I'll take a look.

cesarLima1 (approved):

PR #434, opened 2024-08-13
[TM-1182] create a new version when edit polygon

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1182)


src/components/elements/Map-mapbox/Map.tsx
```diff
@@ -257,22 +262,56 @@ export const MapContainer = ({
...
       if (geojson) {
         if (polygonFromMap?.uuid) {
+          !pdView && onCancelEdit();
           const feature = geojson.features[0];
           const response = await fetchPutV2TerrafundPolygonUuid({
-            body: { geometry: JSON.stringify(feature) },
+            body: {
+              geometry: JSON.stringify(feature),
+              // @ts-ignore
+              adminUpdate: pdView ? false : true
```
pachonjcl, line 288: pdView is already a boolean right? so no need to check

src/components/elements/Map-mapbox/Map.tsx
```diff
@@ -257,22 +262,56 @@ export const MapContainer = ({
...
+            },
             pathParams: { uuid: polygonFromMap?.uuid }
           });
           reloadSiteData?.();
-          if (response.message == "Geometry updated successfully.") {
+          if (
+            response.message == "Site polygon version created successfully." ||
+            response.message == "Geometry updated successfully."
```
pachonjcl, line 295: Please modify backend response, use status code 201 when an entity is created and 200 when an entity is updated.
Later come here and use those status codes for the conditional.
Right now it is depending in synchronizing the messages between BE and FE

pachonjcl (changes requested):

src/components/elements/Map-mapbox/components/OverviewMapArea.tsx
```diff
@@ -205,6 +205,7 @@ const OverviewMapArea = ({
         polygonFromMap={polygonFromMap}
         shouldBboxZoom={!shouldRefetchPolygonData}
         modelFilesData={modelFilesData?.data}
+        pdView={true}
```
cesarLima1, line 208: Just some question:
Did you test this with users who are not Monitoring Partners? What should happen with them? Are they able to create versions?

LimberHope: If you are not a project monitoring partner, you should not have access to the project or those functions.

cesarLima1 (dismissed):

src/components/elements/Map-mapbox/Map.tsx
roguenet: Why do we have an endpoint that can both create and update? These are separate REST actions and should be separate endpoints.

roguenet (approved):

src/components/elements/Map-mapbox/Map.tsx
```diff
@@ -257,23 +263,59 @@ export const MapContainer = ({
...
+      return;
+    }
+    map.current.fitBounds(bounds, {
+      padding: 100,
+      linear: false
+    });
+  };
+
+  const { mutate: putMutate } = usePutV2TerrafundPolygonUuid();
+  const { mutate: postMutate } = usePostV2GeometryUUIDNewVersion();
```
pachonjcl, line 279: rename this two, I suggest `updateGeometry` and `createGeometry`, but up to you, I just find very hard to understand the put and post mutate

pachonjcl (changes requested):

pachonjcl (approved):

PR #513, opened 2024-09-27
[TM-1312] v3 users/me

roguenet:
https://gfw.atlassian.net/browse/TM-1312
[Node PR](https://github.com/wri/terramatch-microservices/pull/4)

Reminder: this won't pass build on Vercel until we have the AWS integration up and running so that API integration build is functional in the lower environments. I'll be working on that in my next ticket.


src/components/generic/Navbar/navbarItems.ts
```diff
@@ -15,10 +15,10 @@ interface INavbarItems {
...
 }
 
-export const getNavbarItems = (t: typeof useT, myOrg?: V2MonitoringOrganisationRead | null): INavbarItems => {
-  const visibility = Boolean(
-    myOrg && myOrg?.status !== "rejected" && myOrg?.status !== "draft" && myOrg.users_status !== "requested"
-  );
+export const getNavbarItems = (t: typeof useT, myOrg?: MyOrganisationConnection): INavbarItems => {
+  const { userStatus, organisation } = myOrg ?? {};
+  const { status } = organisation ?? {};
+  const visibility = Boolean(organisation && status !== "rejected" && status !== "draft" && userStatus !== "requested");
```
cesarLima1, line 21: It could be a good idea to create a separate file to store all the constants for terms like `rejected`, `requested`, etc., since I see they are repeated across components. That way, if we need to change one in the future, we can just update it in one place instead of modifying multiple components.

roguenet: In principal, I agree, at least for related components. For instance, all entities and their reports use the same underlying state machine, so their statuses are the same and if any of them changed, they would all change. These two are unique; organisation status would not necessarily change if the project status strings changed, so unifying them could actually cause the opposite effect of needing to break apart statuses that were kind of arbitrarily combined. 

In this case, I think that creating constants would just be to avoid "magic" strings, which is something that I used to do religiously in JS. However, the TS compiler will make sure we're only using strings that it believes are valid. 

In this case, `userStatus` is coming from the Organisation connection:
```
type OrganisationConnection = {
  organisation?: OrganisationDto;
};

type UserStatus = "approved" | "rejected" | "requested";
export type MyOrganisationConnection = OrganisationConnection & {
  organisationId?: string;
  userStatus?: UserStatus;
};
```

And the Organisation status is defined by the BE via the typings on the OrganisationDto
```
export type OrganisationDto = {
  status: "draft" | "pending" | "approved" | "rejected";
  name: string | null;
};
```

I imagine VS code does this as well, but in WebStorm, I get super handy hints about this stuff:
image
image

roguenet: And here's an illustration of what happens in TS if we tried to compare against a misspelled string value: 
image

cesarLima1:
To avoid build errors in Vercel, merge the `yarn.lock` from staging to prevent using` http://registry.chelsea-apps.com`, this will likely fix the checks.

src/components/generic/Navbar/navbarItems.ts
cesarLima1: In this case, it's best to leave it as is. I will approve it then.

pachonjcl (approved):
Code looks good to me

roguenet:
> To avoid build errors in Vercel, merge the `yarn.lock` from staging to prevent using` http://registry.chelsea-apps.com`, this will likely fix the checks.

yeah, there are a few changes from staging that need to get merged in. I'll do that on the epic branch when it's ready to go. That would fix some problems, but this won't build until I can get the vercel build pointed at an env that has the API gateway and new v3 services in place because of the `generate:api` step.

PR #541, opened 2024-10-08
Feat/tm 1314 dashboard web design shell for the country dashboard

dottyy:
[Task link](https://gfw.atlassian.net/jira/software/c/projects/TM/boards/40?selectedIssue=TM-1314)


diego-morales-flores-1996 (approved):

src/pages/dashboard/components/ContentOverviewProject.tsx
```diff
@@ -145,7 +145,7 @@ const ContentOverviewProject = () => {
               column.header === "Hectares" ? (column.header = "Restoration Hectares") : column.header;
               return {
                 ...column,
-                enableSorting: true
+                enableSorting: column.header?.length === 0 ? false : true
```
cesarLima1, line 148: Instead of checking if `column.header?.length === 0` and returning `false`, you can directly use the logical negation. something like this:
 `enableSorting: Boolean(column.header?.length)`

src/components/generic/Layout/DashboardLayout.tsx
```diff
@@ -1,20 +1,78 @@
...
+    label: string;
+    icon: string;
+  };
+  id: number;
+  country_slug: string;
 }
 
 const DashboardLayout = (props: PropsWithChildren<DashboardLayoutProps>) => {
+  const router = useRouter();
+  const { data: dashboardCountries } = useGetV2DashboardCountries<any>({
```
cesarLima1, line 33: you can replace `any` with a type that better reflects the structure of the data. Assuming `dashboardCountries` is an object with a `data` property that is an array of `CountriesProps`, you can type it like this:
```
const { data: dashboardCountries } = useGetV2DashboardCountries<{ data: CountriesProps[] }>({
  queryParams: {}
});
```
I'm not sure what the structure is, so it should be review.

diego-morales-flores-1996: done

PR #555, opened 2024-10-14
[TM-1322] Implement Workdays For HBF Reporting

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1322

No HBF
Screenshot 2024-10-14 at 12 28 13

HBF
Screenshot 2024-10-14 at 12 17 23

### DIFF
--- a/src/components/extensive/WorkdayCollapseGrid/WorkdayCollapseGrid.tsx
+++ b/src/components/extensive/WorkdayCollapseGrid/WorkdayCollapseGrid.tsx
@@ -23,7 +23,7 @@ const WorkdayCollapseGrid: FC<WorkdayCollapseGridProps> = ({ title, demographics
   const [open, setOpen] = useState(false);
   const t = useT();
   const { framework } = useFrameworkContext();
-  const { total, status } = useTableStatus(framework, demographics);
+  const { total, status } = useTableStatus(demographics);
   const byType = useMemo(() => groupBy(demographics, "type"), [demographics]);
 
   const onSectionChange = useCallback(
@@ -88,7 +88,6 @@ const WorkdayCollapseGrid: FC<WorkdayCollapseGridProps> = ({ title, demographics
           >
             {demographicTypes.map(type => (
               <WorkdaySection
-                framework={framework}
                 key={type}
                 onChange={onChange == null ? undefined : demographics => onSectionChange(type, demographics)}
                 demographics={byType[type] ?? []}
--- a/src/components/extensive/WorkdayCollapseGrid/WorkdayCollapseGrid.tsx
+++ b/src/components/extensive/WorkdayCollapseGrid/WorkdayCollapseGrid.tsx
@@ -5,26 +5,40 @@ import { FC, useCallback, useMemo, useState } from "react";
 import { When } from "react-if";
 
 import Text from "@/components/elements/Text/Text";
+import { Framework, useFrameworkContext } from "@/context/framework.provider";
 
 import Icon, { IconNames } from "../Icon/Icon";
 import { useTableStatus } from "./hooks";
-import { Demographic, DEMOGRAPHIC_TYPES, DemographicType, WorkdayCollapseGridProps } from "./types";
+import {
+  Demographic,
+  DEMOGRAPHIC_TYPES,
+  DemographicType,
+  HBF_DEMOGRAPHIC_TYPES,
+  HBFDemographicType,
+  WorkdayCollapseGridProps
+} from "./types";
 import WorkdaySection from "./WorkdaySection";
 
 const WorkdayCollapseGrid: FC<WorkdayCollapseGridProps> = ({ title, demographics, variant, onChange }) => {
   const [open, setOpen] = useState(false);
   const t = useT();
-  const { total, status } = useTableStatus(demographics);
+  const { framework } = useFrameworkContext();
+  const { total, status } = useTableStatus(framework, demographics);
   const byType = useMemo(() => groupBy(demographics, "type"), [demographics]);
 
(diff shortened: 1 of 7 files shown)


src/components/extensive/WorkdayCollapseGrid/WorkdayCollapseGrid.tsx
```diff
@@ -5,20 +5,29 @@ import { FC, useCallback, useMemo, useState } from "react";
...
+  WorkdayCollapseGridProps
+} from "./types";
 import WorkdaySection from "./WorkdaySection";
 
 const WorkdayCollapseGrid: FC<WorkdayCollapseGridProps> = ({ title, demographics, variant, onChange }) => {
   const [open, setOpen] = useState(false);
   const t = useT();
-  const { total, status } = useTableStatus(demographics);
+  const { framework } = useFrameworkContext();
+  const { total, status } = useTableStatus(framework, demographics);
```
roguenet, line 26: There's no need to pass the framework in here. The `useTableStatus` hook can use `useFrameworkContext` internally.

src/components/extensive/WorkdayCollapseGrid/WorkdaySection.tsx
```diff
@@ -5,23 +5,32 @@ import { When } from "react-if";
...
   demographics: Demographic[];
-  type: DemographicType;
+  type: DemographicType | HBFDemographicType;
   variant: WorkdayGridVariantProps;
+  framework: Framework;
   onChange?: (demographics: Demographic[]) => void;
 }
 
-const WorkdaySection = ({ demographics, type, variant, onChange }: WorkdaySectionProps) => {
+const WorkdaySection = ({ demographics, type, variant, onChange, framework }: WorkdaySectionProps) => {
```
roguenet, line 28: Likewise, please don't pass new props in for things that are supplied by context. This component should use `useFrameworkContext` internally instead.

src/components/extensive/WorkdayCollapseGrid/hooks.ts
```diff
@@ -19,29 +30,29 @@ export function calculateTotals(demographics: Demographic[]) {
...
-    { gender: 0, age: 0, ethnicity: 0 }
+    { gender: 0, age: 0, ethnicity: 0, caste: 0 }
   );
 
-  const total = Math.max(counts.age, counts.gender, counts.ethnicity);
-  const countsMatch = uniq([counts.age, counts.gender, counts.ethnicity]).length === 1;
-  return { total, countsMatch };
+  const total = Math.max(counts.age, counts.gender, counts.ethnicity, counts.caste);
+  const countsMatch = uniq([counts.age, counts.gender, counts.ethnicity, counts.caste]).length === 1;
+  return { counts, total, countsMatch };
```
roguenet, line 38: This is going to break for the existing usage of the workday table. Since non-HBF uses don't have `caste`, it's always going to be 0, and will therefore show as "not balanced" in the editor because `countsMatch` will always be false. Once you have a new strategy in place here, please test with non-HBF frameworks to make sure the existing functionality isn't impacted.

src/components/extensive/WorkdayCollapseGrid/hooks.ts
```diff
@@ -70,19 +81,37 @@ function mapRows(usesSubtype: boolean, typeMap: Dictionary<string>, demographics
...
+function getDemographicTypes<T extends Framework>(framework: T): Readonly<string[]> {
+  return framework === Framework.HBF ? HBF_DEMOGRAPHIC_TYPES : DEMOGRAPHIC_TYPES;
+}
+
+function getDemographicTypesMap<T extends Framework>(framework: T) {
+  return framework === Framework.HBF ? HBF_DEMOGRAPHIC_TYPE_MAP : DEMOGRAPHIC_TYPE_MAP;
+}
+
+export function useSectionData<T extends Framework>(
+  framework: T,
```
roguenet, line 97: use `useFrameworkContext` instead of passing it in.

src/hooks/useWorkdayData.tsx
```diff
@@ -45,17 +52,20 @@ export default function useWorkdayData(
...
-            ) ?? counts
-          );
-        },
-        { gender: 0, age: 0, ethnicity: 0 }
-      );
+      const initialCounts: DemographicCounts = {
+        gender: 0,
+        age: 0,
+        ethnicity: 0,
+        caste: 0
```
roguenet, line 59: We want to avoid sending `caste` at all for non-HBF frameworks, and avoid sending `ethnicity` for HBF frameworks, so this needs to be framework-aware as well.

src/components/extensive/WorkdayCollapseGrid/hooks.ts
```diff
@@ -14,23 +62,35 @@ export interface SectionRow {
...
+    return counts;
+  }, initialCounts);
+
+  let total: number = 0;
+  let countsMatch: boolean = false;
+
+  if (counts) {
+    if (isHBFDemographicCounts(counts, framework)) {
+      total = Math.max(counts.age, counts.gender, counts.caste);
+      countsMatch = uniq([counts.age, counts.gender, counts.caste]).length === 1;
```
roguenet, line 79: I think this should just be `true` or maybe `counts.gender > 0`. The status in `useTableStatus` looks like it'll never by "complete" for HBF. I think I'd make these changes:
* Rename `countsMatch` to `complete` to make it a bit more generic, and set it to `counts.gender > 0` in this block
* Set `total` in this block to `counts.gender` and then the change to the return value of `useTableStatus` below can be reverted.

roguenet (dismissed):

pachonjcl (approved):

PR #583, opened 2024-10-22
[TM-1408] helper text missing on file

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1408)


roguenet (commented):
@LimberHope If this is going to be based on the release branch, then your source branch needs to be cut from the release branch. This has a bunch of stuff on it from staging that should not be here.

PR #603, opened 2024-10-30
[TM-1415] HBF Workday Implementation corrections

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1415

### DIFF
--- a/src/components/extensive/WorkdayCollapseGrid/types.ts
+++ b/src/components/extensive/WorkdayCollapseGrid/types.ts
@@ -50,13 +50,15 @@ const CASTES: Dictionary<string> = {
 };
 
 const AGES: Dictionary<string> = {
-  youth: "Youth (15-29)",
-  adult: "Adult (29-64)",
+  youth: "Youth (15-24)",
+  adult: "Adult (24-64)",
   elder: "Elder (65+)",
   unknown: "Unknown"
 };
 
-const HBF_AGES = Object.fromEntries(Object.entries(AGES).filter(([key]) => key === "youth"));
+const HBF_AGES: Dictionary<string> = {
+  youth: "Youth (15-29)"
+};
 
 const ETHNICITIES: Dictionary<string> = {
   indigenous: "Indigenous",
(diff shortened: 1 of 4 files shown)


src/components/extensive/WorkdayCollapseGrid/types.ts
```diff
@@ -50,8 +50,8 @@ const CASTES: Dictionary<string> = {
 };
 
 const AGES: Dictionary<string> = {
-  youth: "Youth (15-24)",
-  adult: "Adult (24-64)",
+  youth: "Youth (15-29)",
+  adult: "Adult (29-64)",
```
roguenet, line 54: This change is affecting non-HBF values as well. I believe Claire is asking for the value to change for HBF only.

src/helpers/customForms.ts
```diff
@@ -644,7 +644,7 @@ const getFieldValidation = (question: FormQuestionRead, t: typeof useT, framewor
             const { demographics } = value.length > 0 ? value[0] : {};
             if (demographics == null) return true;
 
-            return calculateTotals(demographics, framework).complete;
+            return calculateTotals(demographics, framework).counts;
```
roguenet, line 647: I don't believe this is correct. this `test` method is supposed to return a boolean indicating if the error message should display or not. I updated this line earlier today and I'm pretty sure `complete` is correct.

FWIW, this line as it was this morning is probably what was causing the error with save & continue mentioned in the ticket.

roguenet:
Liza was wanting this to go into the release, so I've tagged the ticket and I'm moving this PR to the release branch. Also, next time use a name for the branch more like `fix/TM-1415-workday-corrections` per our branch naming standards.

src/components/extensive/WorkdayCollapseGrid/types.ts
```diff
@@ -56,7 +56,9 @@ const AGES: Dictionary<string> = {
   unknown: "Unknown"
 };
 
-const HBF_AGES = Object.fromEntries(Object.entries(AGES).filter(([key]) => key === "youth"));
+const HBF_AGES: Dictionary<string> = {
```
roguenet, line 59: Should there be something new referencing this?

roguenet (approved):

PR #653, opened 2024-11-11
[TM-572] "Nothing to report" selection bumps user to page 1 of reports

Scriptmatico:
https://gfw.atlassian.net/browse/TM-572

The issue was that the hook useDate is returning a reference everytime any state is updated, this for useMemo is change even if the other data haven't changed this reload the whole page.

Now memoizing the format function in the useDate will cache the function and just change when is needed.


roguenet (approved):
Good catch!

PR #687, opened 2024-11-19
[TM-1489] Update middleware to handle dashboard access for all user t…

cesarLima1:
https://gfw.atlassian.net/browse/TM-1489

### DIFF
--- a/src/middleware.page.ts
+++ b/src/middleware.page.ts
@@ -19,6 +19,14 @@ export async function middleware(request: NextRequest) {
       return NextResponse.next();
     }
 
+    if (accessToken && middlewareCache) {
+      // Specific routes handling for cached middleware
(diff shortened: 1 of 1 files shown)


roguenet (approved):
I don't fully understand this whole refactor, and this is a strong regression risk. Let's make sure Rod does a pretty thorough regression pass on typical user flows, making sure that if the user does a refresh on various pages that they don't get redirected somewhere unexpected.

cesarLima1:
The new version introduces the next changes in general terms:
1. Adds explicit handling for the /dashboard route, ensuring it works as intended.
2. Change middleware cache logic with control over route-specific redirects.
3. Add role-based routing, especially for funder/government users.
As you mention it needs to be fully tested to avoid another hotfix.

PR #703, opened 2024-11-26
[TM-1513] add attemps for delayed jobs

egrojMonroy:
After a 502 on getting job status, try 3 times .

### DIFF
--- a/src/generated/apiFetcher.ts
+++ b/src/generated/apiFetcher.ts
@@ -136,7 +136,7 @@ type JobResult = {
   };
 };
 
-async function loadJob(signal: AbortSignal | undefined, delayedJobId: string): Promise<JobResult> {
+async function loadJob(signal: AbortSignal | undefined, delayedJobId: string, retries: number = 3): Promise<JobResult> {
   let response, error;
   try {
     const headers: HeadersInit = { "Content-Type": "application/json" };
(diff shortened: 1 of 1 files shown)


src/generated/apiFetcher.ts
```diff
@@ -145,6 +145,12 @@ async function loadJob(signal: AbortSignal | undefined, delayedJobId: string): P
 
     const url = resolveV3Url(`/jobs/v3/delayedJobs/${delayedJobId}`);
     response = await fetch(url, { signal, headers });
+
+    if (response.status === 502 && retries > 0) {
```
roguenet, line 149: Please add a comment explaining why we're doing this; if I were a developer coming across this code in 6 months, I'd be pretty confused about why this is here.

src/generated/apiFetcher.ts
```diff
@@ -136,7 +136,7 @@ type JobResult = {
   };
 };
 
-async function loadJob(signal: AbortSignal | undefined, delayedJobId: string): Promise<JobResult> {
+async function loadJob(signal: AbortSignal | undefined, delayedJobId: string, retries: number = 3): Promise<JobResult> {
```
roguenet, line 139: 🎨 There's no reason to specify the type when a default value is provided; TS compiler will do just fine understanding it.

src/generated/apiFetcher.ts
egrojMonroy: Great, a comment is added.

src/generated/apiFetcher.ts
egrojMonroy: removed

roguenet (approved):

PR #745, opened 2024-12-13
[TM-1531] add job connection

egrojMonroy:
add the connection to pull the list of jobs


src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,57 @@
...
+    pathParams: { uuid }
+  };
+
+  return delayedJobsFindFetchFailed(variables)(store) != null;
+};
+const connectionIsLoaded = ({ delayedJobs, hasLoadFailed, isLoading }: DelayedJobsConnection) => {
+  return (delayedJobs != null && delayedJobs.length > 0) || hasLoadFailed || isLoading;
+};
+
+const delayedJobsConnection: Connection<DelayedJobsConnection, DelayedJobsProps> = {
```
roguenet, line 35: These props aren't used, no reason to include them.

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const delayedJobsSelector = (store: ApiDataStore) => {
+  const delayedJobsMap = store.delayedJobs || {};
+  return Object.values(delayedJobsMap).map(resource => resource.attributes);
+};
+
+const delayedJobsLoadFailedSelector = (store: ApiDataStore, { uuid }: DelayedJobsProps) => {
+  const variables: DelayedJobsFindVariables = {
+    pathParams: { uuid }
+  };
```
roguenet, line 27: Since the API request doesn't use these path params, the load failed selector shouldn't either. The params to those two must always match for this feature to function.

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,57 @@
...
+    const isLoaded = connectionIsLoaded(connection);
+    if (!isLoaded) {
+      listDelayedJobs();
+    }
+  },
+  isLoaded: connectionIsLoaded,
+  selector: selectorCache(
+    props => props.uuid,
+    props =>
+      createSelector(
```
roguenet, line 46: Since there are no props, the selector cache isn't needed - it's only needed when there are props to differentiate different selectors. The `selector` here can just receive `createSelector`

roguenet (commented):
A quick review pass to hopefully nudge you closer to the right direction.

src/connections/DelayedJob.ts
egrojMonroy: understood and removed

src/connections/DelayedJob.ts
egrojMonroy: removed

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+
+// --- Delayed Jobs Connection ---
+type DelayedJobsConnection = {
+  delayedJobs?: DelayedJobDto[];
+  isLoading: boolean;
+  hasLoadFailed: boolean;
+};
+
+const delayedJobsSelector = (store: ApiDataStore) => {
+  const delayedJobsMap = store.delayedJobs || {};
```
roguenet, line 23: Please use ?? for defaults. I'd reformat this whole method:
```
const delayedJobsSelector = (store: ApiDataStore) => 
  Object.values(store.delayedJobs ?? {}).map(resource => resource.attributes);
```

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+  return Object.values(delayedJobsMap).map(resource => resource.attributes);
+};
+
+const delayedJobsLoadFailedSelector = (store: ApiDataStore) => {
+  return listDelayedJobsFetchFailed(store) != null;
+};
```
roguenet, line 29: Single statement arrow functions should use the shortcut:
```
const delayedJobsLoadFailedSelector = (store: ApiDataStore) => listDelayedJobsFetchFailed(store) != null;
```

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const connectionIsLoaded = ({ delayedJobs, hasLoadFailed, isLoading }: DelayedJobsConnection) => {
+  return (delayedJobs != null && delayedJobs.length > 0) || hasLoadFailed || isLoading;
+};
```
roguenet, line 33: Same here - please use the arrow function shortcut.

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+const delayedJobsConnection: Connection<DelayedJobsConnection> = {
+  load: connection => {
+    const isLoaded = connectionIsLoaded(connection);
+    if (!isLoaded) {
+      listDelayedJobs();
+    }
+  },
+  isLoaded: connectionIsLoaded,
+  selector: createSelector(
+    [delayedJobsSelector, store => delayedJobsLoadFailedSelector(store)],
```
roguenet, line 44: This second argument does not need to create a fresh function. This line can be
```
[delayedJobsSelector, delayedJobsLoadFailedSelector],
```

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+    (delayedJobs, hasLoadFailed): DelayedJobsConnection => ({
```
roguenet, line 45: 🎨 It's OK to leave it, but I believe setting the return type here is not necessary (the type comes from the connection shape... if your return type didn't match DelayedJobsConnection, it would complain even if you don't specify it here):
```
(delayedJobs, hasLoadFailed) => ({
```

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+      delayedJobs,
+      isLoading: delayedJobs == null && !hasLoadFailed,
+      hasLoadFailed
+    })
+  )
+};
+
+export const useDelayedJobs = () => {
+  const connection = connectionHook(delayedJobsConnection)();
```
roguenet, line 54: This should use `useConnection` directly.
```
const connection = useConnection(delayedJobsConnection);
```

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  useEffect(() => {
+    const intervalId = setInterval(() => {
+      listDelayedJobs();
+    }, 1500);
```
roguenet, line 59: This will cause this interval to keep refreshing the delayed jobs for as long as the surrounding component is mounted. Is that intentional? It seems like clearing the interval when all the delayed jobs have completed would be good to keep from continuing to hit the jobs endpoint. I'm OK with letting this live to prod if that's complicated to wire up; it's just not ideal and should be cleaned up in a future ticket.

Please either decide to implement an eventual stop like that in this PR, or see that a tech debt ticket is created to clean up in a future release.

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+    return () => {
+      clearInterval(intervalId);
+    };
+  }, []);
+
+  return connection;
+};
+
+export const triggerBulkUpdate = (jobs: DelayedJobData[]) => {
+  return bulkUpdateJobs({ body: { data: jobs } });
```
roguenet, line 70: I don't think there's any reason to return here. Functions that trigger v3 API calls should never be reliant on their return value; the data must always come from the redux store.

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+};
+
+const bulkUpdateJobsSelector = (store: ApiDataStore) => {
+  const bulkUpdateState = store.delayedJobs || {};
+  return {
+    isLoading: bulkUpdateJobsIsFetching(store),
+    hasLoadFailed: bulkUpdateJobsFetchFailed(store) != null,
+    response: bulkUpdateState.response
+  };
+};
```
roguenet, line 80: This should use `??` for defaults, although I think this is an overall better format for this method:
```
const bulkUpdateJobsSelector = (store: ApiDataStore) => ({
  isLoading: bulkUpdateJobsIsFetching(store),
  hasLoadFailed: bulkUpdateJobsFetchFailed(store) != null,
  response: store.delayedJobs?.reponse
});
```

However, I'm also confused as to why you would expect a `response` object to be on `store.delayedJobs`. That shouldn't be possible with the current redux setup.

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const bulkUpdateJobsConnection: Connection<DelayedJobsConnection, { jobs: DelayedJobData[] }> = {
+  load: async (connection, { jobs }) => {
+    const isLoaded = connectionBulkUpdateIsLoaded(connection);
+    if (!isLoaded) {
+      await bulkUpdateJobs({ body: { data: jobs } });
```
roguenet, line 86: connections should never have await / async mechanics. The data will be loaded into the redux store when it's been fetched, and then isLoaded / selector are called again.

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  },
+
+  isLoaded: state => !state.isLoading,
+
+  selector: createSelector([store => bulkUpdateJobsSelector(store)], ({ isLoading, hasLoadFailed, response }) => ({
```
roguenet, line 92: Please avoid creating extraneous intermediate functions:
```
selector: createSelector([bulkUpdateJobsSelector], ({
```

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,103 @@
...
+    isLoading,
+    hasLoadFailed,
+    response
+  }))
+};
+
+const connectionBulkUpdateIsLoaded = ({ isLoading, hasLoadFailed }: { isLoading: boolean; hasLoadFailed: boolean }) => {
+  return !isLoading && !hasLoadFailed;
+};
```
roguenet, line 101: Please use the arrow function shortcut:
```
=> !isLoading && !hasLoadFailed;
```

src/store/apiSlice.ts
```diff
@@ -45,6 +45,7 @@ export type Relationships = {
 };
 
 export type StoreResource<AttributeType> = {
+  id?: string;
```
roguenet, line 48: Why was this added?

src/connections/DelayedJob.ts
egrojMonroy: changed

src/connections/DelayedJob.ts
egrojMonroy: changed

src/connections/DelayedJob.ts
egrojMonroy: changed

src/connections/DelayedJob.ts
egrojMonroy: understood and changed

src/connections/DelayedJob.ts
egrojMonroy: improved

src/connections/DelayedJob.ts
egrojMonroy: done

src/connections/DelayedJob.ts
egrojMonroy: understood, and changed

src/connections/DelayedJob.ts
egrojMonroy: removed

src/connections/DelayedJob.ts
egrojMonroy: changed

src/connections/DelayedJob.ts
egrojMonroy: changed

src/store/apiSlice.ts
egrojMonroy: this should have not reached git. It was from an understanding of the store. Removed

src/connections/DelayedJob.ts
egrojMonroy: the response object wasnt needed, and everything keeps working.

src/connections/DelayedJob.ts
egrojMonroy: Yes indeed, it is supposed to be attentive for any job created on any session. 
Nonetheless I think it might not be good. 
So for now the debt ticket will be needed, because we would have to think on the user cases.

src/connections/DelayedJob.ts
```diff
@@ -66,38 +59,32 @@ export const useDelayedJobs = () => {
...
     }
   },
 
   isLoaded: state => !state.isLoading,
 
-  selector: createSelector([store => bulkUpdateJobsSelector(store)], ({ isLoading, hasLoadFailed, response }) => ({
+  selector: createSelector(bulkUpdateJobsSelector, ({ isLoading, hasLoadFailed, response }) => ({
     isLoading,
     hasLoadFailed,
     response
```
roguenet, line 83: This seems like it should be throwing an error in the TS compiler... Maybe I'm missing something, but this shape does not seem to match `DelayedJobsConnection`. 

A couple of thoughts here:
* Typically, each connection will have its own shape. That's not required, but it's normal
* I've found in the past that properties like `isLoading`, `hasLoadFailed` and `response`, while seeming convenient when looking at a single connection, are actually a disaster down the road. The reason being that if we follow such generic conventions for individual connections, then when a component ends up using 2 or more connections, it gets really awkward to juggle all the `isLoading`s that each connection returns. Instead, I think it's better to be super specific with the properties of the connection shape. So something more like `bulkUpdateJobsIsLoading` and `bulkUpdateJobsHasFailed`.

As for `response` - this appears to be just returning all the delayed jobs like the other connection does. Are these used in the same component? Perhaps a better approach would be to combine them into a single connection. I think of connections as being responsible for providing a given set of data from the redux cache, and a set of functions for operating on that data. It would be totally reasonable for this `bulkUpdateJobs` call to be triggered from a function that's included on the other connection. Or, the bulkUpdate function could be simply exposed by this file, and then we simply rely on the original connection to provide the `delayedJobs` from the store. One of the primary wins of this system is that a connection returns data that got cached in the store _regardless of where it came from_. So if one call does the initial fetch, and another call modifies that data, as long as the data ends up back in the store under the same ID, the component that relies on it will rerender when it gets updated.

egrojMonroy: Finally joined both connections. 
Tested and is working.

src/components/elements/Notification/FloatNotification.tsx
```diff
@@ -38,36 +83,81 @@ const FloatNotification = ({ data }: FloatNotificationProps) => {
...
+                                  : ((item.processedContent ?? 0) / (item.totalContent ?? 1)) * 100
+                              }
+                              className="h-2 bg-success-40"
+                              color="success-600"
+                            />
+                          )}
+                          <Text variant="text-12-semibold" className="text-black">
+                            {item.name === "Polygon Upload"
+                              ? item.status === "succeeded"
+                                ? "Done!"
```
roguenet, line 138: Please make sure any text that is displayed gets wrapped in `t()`

src/connections/DelayedJob.ts
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const delayedJobsSelector = (store: ApiDataStore) =>
+  Object.values(store.delayedJobs ?? {}).map(resource => resource.attributes);
+
+const bulkUpdateJobsSelector = (store: ApiDataStore) => ({
+  bulkUpdateJobsIsLoading: bulkUpdateJobsIsFetching(store),
+  bulkUpdateJobsHasFailed: bulkUpdateJobsFetchFailed(store) != null,
+  updatedJobsResponse: Object.values(store.delayedJobs ?? {}).map(resource => resource.attributes as DelayedJobDto)
+});
```
roguenet, line 27: These two selectors are problematic because they're inputs to `createSelector`, and create a new object every time they're called. If their inputs don't change, they should return the same object every time to gain the true benefit of reselect. 

A better approach is for the delayedJobsSelector to be simpler: 
```
const delayedJobsSelector = (store: ApiDataStore) => store.delayedJobs
```

and to remove this bulk update selecot, instead relying directly on the isFetching / isFailed selectors in the `createSelector` call. So it becomes something more like:
```
const combinedSelector = createSelector([
  delayedJobsSelector, bulkUpdateJobsIsFetching, bulkUpdateJobsFetchFailed
], (delayedJobs, bulkUpdateJobsIsLoading, bulkUpdateJobsFailure) => ({
  delayedJobs: Object.values(delayedJobs ?? {}).map(resource => resource.attributes),
  ... // other members of the connection
}));
```

It's important for the members of the first argument array to `createSelector` to only return a unique object result if the inputs have changed, whenever possible.

src/components/elements/Notification/FloatNotification.tsx
egrojMonroy: added to every place but names

src/connections/DelayedJob.ts
egrojMonroy: changed it and is working still.

roguenet (approved):

PR #820, opened 2025-01-22
[TM-1311] Implement reset password for MS

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1311


src/pages/auth/reset-password/[token].page.tsx
```diff
@@ -7,7 +7,8 @@ import * as yup from "yup";
 import { usePasswordStrength } from "@/components/extensive/PasswordStrength/hooks/usePasswordStrength";
 import BackgroundLayout from "@/components/generic/Layout/BackgroundLayout";
 import ContentLayout from "@/components/generic/Layout/ContentLayout";
-import { usePatchAuthChange } from "@/generated/apiComponents";
+// import { usePatchAuthChange } from "@/generated/apiComponents";
```
Comment (roguenet) [inline src/pages/auth/reset-password/[token].page.tsx:10]: Please don't leave in commented out code.

src/pages/auth/reset-password/[token].page.tsx
```diff
@@ -44,10 +43,12 @@ const ResetPasswordPage = () => {
       if (!router.query.token) {
         router.push("/");
       } else {
-        await requestResetPassword({
+        NewResetPassword({
```
Comment (roguenet) [inline src/pages/auth/reset-password/[token].page.tsx:46]: The generated API methods in v3 are not meant to be used directly in components. Please read up on the connection system and see its uses in this codebase: https://gfw.atlassian.net/wiki/spaces/TerraMatch/pages/1423147024/Connections

src/store/apiSlice.ts
```diff
@@ -311,3 +320,53 @@ export default class ApiSlice {
     this.redux.dispatch(apiSlice.actions.setAbortDelayedJob(abort_delayed_job));
   }
 }
+
+export const selectResetPassword = (store: ApiDataStore) => Object.values(store.logins)?.[0]?.attributes;
+
+type RequestResetPasswordConnection = {
```
roguenet, line 326: Connections should be in their own file in `src/connections`, not in the apiSlice file.

src/connections/ResetPassword.ts
```diff
@@ -0,0 +1,61 @@
...
+  requestPasswordResetFetchFailed,
+  requestPasswordResetIsFetching,
+  resetPasswordFetchFailed,
+  resetPasswordIsFetching
+} from "@/generated/v3/userService/userServicePredicates";
+import { ApiDataStore, PendingErrorState } from "@/store/apiSlice";
+import { Connection } from "@/types/connection";
+import { connectionHook } from "@/utils/connectionShortcuts";
+
+export const selectResetPassword = (store: ApiDataStore) => Object.values(store.logins)?.[0]?.attributes;
```
roguenet, line 13: As noted in the microservices PR, this resource should not be getting stored under a different resource type.

src/connections/ResetPassword.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+const requestPasswordConnection: Connection<RequestResetPasswordConnection> = {
+  selector: createSelector(
+    [requestPasswordResetIsFetching, requestPasswordResetFetchFailed, selectResetPassword],
+    (isLoading, requestFailed, selector) => {
+      return {
+        isLoading: isLoading,
+        requestFailed: requestFailed,
+        isSuccess: (selector as any) != null,
```
roguenet, line 29: Why is `as any` needed here? The FE codebase isn't very well typed in a lot of cases so some `any`s are to be expected, but I've put a lot of work into making sure we can start to fix that with v3. I wouldn't expect an `any` anywhere in Connection code.

src/connections/ResetPassword.ts
```diff
@@ -0,0 +1,61 @@
...
+  isLoading: boolean;
+  requestFailed: PendingErrorState | null;
+  isSuccess: boolean;
+};
+
+const resetPasswordConnection: Connection<ResetPasswordConnection> = {
+  selector: createSelector(
+    [
+      resetPasswordIsFetching({ pathParams: { token: "" } }),
+      resetPasswordFetchFailed({ pathParams: { token: "" } }),
```
roguenet, line 46: This isn't going to work correctly. The token needs to be included as a param on the connection. These two predicates work by matching against the full URL of the request against the request tracking in the store, so for them to work correctly, the path params need to be accurate.

src/pages/auth/reset-password/[token].page.tsx
```diff
@@ -44,10 +45,12 @@ const ResetPasswordPage = () => {
       if (!router.query.token) {
         router.push("/");
       } else {
-        await requestResetPassword({
+        resetPassword({
```
Comment (roguenet) [inline src/pages/auth/reset-password/[token].page.tsx:48]: Please don't call v3 API methods directly from components. The ResetPasswordConnection should accept `token` as a prop, and expose a `resetMethod` function that just takes the new password and knows how to apply the password to the body and the token to the URL. Component code shouldn't have to know anything about how things are being updated, but should simply have access to methods for modifying data while the Connection code handles the specifics.

As mentioned above, the connection needs the `token` as a prop anyway for `isLoading` and `requestFailed` to work correctly.

src/pages/auth/reset-password/confirm.page.tsx
```diff
@@ -23,16 +25,18 @@ const SignupConfirmPage = () => {
...
+    if (isSuccess) {
       timer.reset();
       timer.start();
     }
   });
 
   const handleResend = async () =>
-    requestResetPassword({
-      body: { email_address: email, callback_url: window.location.origin + `${baseAuthPath}/reset-password/` }
+    requestPasswordReset({
```
roguenet, line 38: This method from the connection should simply accept the email address and callback URL directly. The component code doesn't need to know how they're going to be used, that's the job of the Connection.

In general, in v3:
 * Components declare what their requirements are by using connections. They are just trying to say "I need a method to trigger a password reset, and I need to know if it worked"
 * Connections are responsible for providing the requirements that a component might need, and part of that is providing methods that interface with the API layer. 
 
Components should ideally not know or care about two things: where the data they require came from, and what the particulars are of updating those data.

src/connections/ResetPassword.ts
Scriptmatico: Im not sure what do you mean here, can you explain a little more? Or even if you can provide an example will be great!

roguenet: Take a look at the code for those generated predicates - if you dig enough, you'll get to `isFetching` in [utils.ts](https://github.com/wri/wri-terramatch-website/blob/6351508b8694c2f63317d33e613f5f34915ce938/src/generated/v3/utils.ts#L58). That method shows that the variables that are used to generate the URL for the request are important because in the ApiStore, we track the state of requests based on URL (take a look at the [reducers in apiSlice.ts](https://github.com/wri/wri-terramatch-website/blob/ba0ce14f23fb1787d6c748e55485e1da372565ef/src/store/apiSlice.ts#L156).

So, in order for these two predicates to work, the token that was used for the request has to be included. You'll need to refactor this connection to something like this:

```
type ResetPasswordConnection = {
  isLoading: boolean;
  requestFailed: PendingErrorState | null;
  isSuccess: boolean;
  resetPassword: (password: string) => void;
};

type ResetPasswordProps = {
  token: string;
};

const resetPasswordConnection: Connection<ResetPasswordConnection, ResetPasswordProps> = {
  selector: selectorCache(
    ({ token }) => token,
    ({ token }) =>
      createSelector(
        [
          resetPasswordIsFetching({ pathParams: { token } }),
          resetPasswordFetchFailed({ pathParams: { token } }),
          selectResetPassword
        ],
        (isLoading, requestFailed, selector) => ({
          isLoading,
          requestFailed,
          isSuccess: selector?.emailAddress != null,

          resetPassword: (password: string) =>
            resetPassword({
              body: {
                newPassword: password
              },
              pathParams: {
                token: token
              }
            })
        })
      )
  )
};
```

With that in place you'd also remove your exported `PasswordReset` function, and use the one from the connection instead. Since the connection already has the token connected to it, the resetPassword function that it can export requires fewer parameters.

roguenet: The [EstablishmentTrees connection](https://github.com/wri/wri-terramatch-website/blob/d2c024218f5626bc040f4176b97240dad67bae1f/src/connections/EstablishmentTrees.ts#L49) is another good place to see this pattern in use.

src/connections/ResetPassword.ts
```diff
@@ -0,0 +1,82 @@
...
+  requestPasswordResetIsFetching,
+  resetPasswordFetchFailed,
+  resetPasswordIsFetching
+} from "@/generated/v3/userService/userServicePredicates";
+import { ApiDataStore, PendingErrorState } from "@/store/apiSlice";
+import { Connection } from "@/types/connection";
+import { connectionHook } from "@/utils/connectionShortcuts";
+import { selectorCache } from "@/utils/selectorCache";
+
+export const RequestPasswordReset = (emailAddress: string, callbackUrl: string) =>
```
roguenet, line 15: exported functions should start with a lower case (`requestPasswordReset`). Starting with upper case is reserved for classes.

roguenet (approved):

PR #900, opened 2025-02-07
[TM-1712] Improve delayed jobs handling

roguenet:
https://gfw.atlassian.net/browse/TM-1712

This does still have a problem in that if nobody ever clears out their delayed jobs notification, it'll keep polling. This will however greatly reduce the amount of delayed job polling that the typical user is doing when they land on the site. 

I'm going to bring in tickets for auto-acknowledging jobs after some reasonable time delay in a future sprint. Will probably start soft deleting them too, we'll see.


src/admin/components/Alerts/DelayedJobsProgressAlert.tsx
```diff
@@ -12,47 +11,16 @@ type DelayedJobsProgressAlertProps = {
...
-      }, 1000);
-    }
-
-    return () => {
-      if (intervalId) {
-        setDelayedJobProcessing(0);
-        setDalayedJobTotal(0);
-        setProgressMessage("Running 0 out of 0 polygons (0%)");
-        clearInterval(intervalId);
-      }
```
roguenet, line 39: I saw this interval to poll the store pattern in a couple of places. One of the great things about redux is that it's easy to set up a selector that only re-draws the current component when its output changes. So, you both hear about store changes immediately, and only redraw when they happen.

See the definition of `useJobProgress()` in `DelayedJob.ts` to see the replacement.

src/admin/components/Alerts/DelayedJobsProgressAlert.tsx
```diff
@@ -12,47 +11,16 @@ type DelayedJobsProgressAlertProps = {
...
-    };
-  }, [show]);
+  const { totalContent, processedContent, progressMessage } = useJobProgress();
 
   const abortDelayedJob = () => {
-    ApiSlice.abortDelayedJob(true);
-    ApiSlice.addTotalContent(0);
-    ApiSlice.addProgressContent(0);
-    ApiSlice.addProgressMessage("Running 0 out of 0 polygons (0%)");
```
roguenet, line 47: This was the other pattern I wanted to call out. Calling several redux store updates in a row is an anti-pattern. Much better to have a single action dispatcher that issues all the changes in a single event. Because of the careful processing of caching selectors, this kind of the thrashing, while not a huge deal, is a performance bottleneck that we'd like to avoid.

src/store/store.tsx
```diff
@@ -4,19 +4,22 @@ import { PropsWithChildren, useMemo } from "react";
...
   api: ApiDataStore;
+  jobs: JobsDataStore;
 };
 
 export const makeStore = (queryClient?: QueryClient) => {
   const store = configureStore({
     reducer: {
-      api: apiSlice.reducer
+      api: apiSlice.reducer,
+      jobs: jobsSlice.reducer
```
roguenet, line 22: As you can see here, it's really easy to create isolated "slices" of the redux store. Next time we have some globally accessible state we want to share between lots of places in the app stack / component tree, spinning up a new slice is greatly preferable to slotting them into an existing one. Nice separation of concerns this way.

src/connections/DelayedJob.ts
```diff
@@ -14,56 +19,76 @@ type DelayedJobCombinedConnection = {
...
-  delayedJobs,
-  delayedJobsHasFailed,
-  bulkUpdateJobsIsLoading,
-  bulkUpdateJobsHasFailed
-}: DelayedJobCombinedConnection) =>
-  (delayedJobs != null || delayedJobsHasFailed) && !bulkUpdateJobsIsLoading && !bulkUpdateJobsHasFailed;
-
 const delayedJobsCombinedConnection: Connection<DelayedJobCombinedConnection> = {
-  load: combinedLoad,
-  isLoaded: combinedIsLoaded,
```
roguenet, line 50: This was an interesting one to track down. These weren't actually doing anything. The reason is that because of the definition of `combinedIsLoaded` (specifically the `delayedJobs != null`), `isLoaded` was always returning true, even with a totally empty store. The reason for that is that each of the resource stores in the api slice starts as an empty object, so there is never a point in time when `store.delayedJobs` is null.

I thought about fixing it up, but realized that if the first call to delayed jobs returns an empty array, it doesn't change the store in any way, so I'd have to implement some sort of a flag in the store or cached in the selectors to indicate whether `delayedJobsIsLoading` has ever been true. An approach like that is valid, and we'll probably end up doing that sooner or later for some connection, but I realized it just wasn't important for this connection. We don't really need to know if it's loaded. We just need to make sure that it gets called at least once on app boot or when a user logs in. See the updated implementation of `useDelayedJobs` to see how that's accomplished. I think this is significantly cleaner than keeping a flag around that just tracks if we've ever called the jobs endpoint.

src/connections/DelayedJob.ts
```diff
@@ -14,56 +19,76 @@ type DelayedJobCombinedConnection = {
...
   selector: combinedSelector
 };
 
+export const useJobProgress = () => useSelector<AppStore, JobsDataStore>(({ jobs }) => jobs);
```
roguenet, line 44: This is literally all that's needed to get immediate updates when a section of the redux store changes.

src/connections/DelayedJob.ts
```diff
@@ -14,56 +19,76 @@ type DelayedJobCombinedConnection = {
...
+      intervalRef.current = undefined;
+    }
+  }, []);
+  const startPolling = useCallback(() => {
+    if (intervalRef.current == null) {
+      intervalRef.current = setInterval(() => {
+        listDelayedJobs();
+      }, 1500);
+    }
+  }, []);
```
roguenet, line 64: These methods are idempotent - you can call them as many times as you like and if their job is already complete, they simply NOOP. That turns out to be really handy for simplifying the code in the effect below that starts and stops polling based on conditions in the redux store.

src/admin/components/Alerts/DelayedJobsProgressAlert.tsx
```diff
@@ -12,47 +11,16 @@ type DelayedJobsProgressAlertProps = {
 };
 
 const DelayedJobsProgressAlert: FC<DelayedJobsProgressAlertProps> = ({ show, title, setIsLoadingDelayedJob }) => {
```
roguenet, line 13: I updated this component, but when I went to look into how to test it I found that it's completely unused. Are there plans to use it in the future, or is this dead code? If it's dead, I think we should go ahead and remove it in this PR. It's also the only place that ever calls `JobSlice.abortDelayedJob`, so if this isn't going to be used, we could rip out the delayed job abort code entirely.

LimberHope (approved):

PR #910, opened 2025-02-08
[TM-1715] React hooks rules

roguenet:
https://gfw.atlassian.net/browse/TM-1715

This is a big refactor. I believe that I kept the intention behind each use of useEffect, useMemo and useCallback the same and improved things along the way.

Two general patterns I noticed:
* Sometimes, a single parameter was passed in the dependencies to `useEffect`, meaning that the effect should only execute if that dep changes. In cases like this, `useValueChanged` will let you avoid the linter problem and make the intention of your code clearer.
* At other times, an empty array was passed to `useEffect` meaning that it should only execute when the component mounts and never again. We have a custom hook for this as well: `useOnMount`. This also makes the intention of the code clearer while avoiding the linter rule.

There were a couple of other patterns I noted, especially in map components, but those ones are generally way too big and complicated to follow for me to extract a good pattern. Mainly just know this: if you're finding that your component is using many instances of useEffect, you're probably in anti-pattern town. Especially if you find yourself setting effects that depend on changing values from useState hooks. This is going to lead to a lot of rerenders and gradual state changes of the overall component that are both bad for performance and bad for user experience. I hope that some day we can revisit the these incredibly complicated mapping components and get them into a better state, but this refactor is already complicated enough; I didn't want to bog it down with further investigations along those lines.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -53,7 +53,7 @@ module.exports = {
       }
     ],
     "no-unused-vars": "off",
-    "react-hooks/exhaustive-deps": "warn",
+    "react-hooks/exhaustive-deps": "error",
     "prettier/prettier": [
       "error",
       {},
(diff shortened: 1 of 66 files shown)


egrojMonroy (approved):
Thanks for the description. Read the code and is a cool learning.

LimberHope (dismissed):

egrojMonroy (approved):

PR #919, opened 2025-02-11
[TM-1735] Fix console error spam on the user list page.

roguenet:
https://gfw.atlassian.net/browse/TM-1735

### DIFF
--- a/src/admin/modules/user/components/UserList.tsx
+++ b/src/admin/modules/user/components/UserList.tsx
@@ -19,7 +19,7 @@ import {
 import { UserDataProvider } from "@/admin/apiProvider/dataProviders/userDataProvider";
 import ListActionsCreateFilter from "@/admin/components/Actions/ListActionsCreateFilter";
 import ExportProcessingAlert from "@/admin/components/Alerts/ExportProcessingAlert";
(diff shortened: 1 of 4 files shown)


src/admin/modules/user/components/UserList.tsx
```diff
@@ -89,7 +91,7 @@ const UserDataGrid = () => {
       <FunctionField
         label="Actions"
         render={(record: any) => {
-          const canEdit = useCanUserEdit(record, "user");
```
roguenet, line 92: This was the source of the copious errors in the JS console. Two things:
* This context does not support hooks - they may only be used in the main body of a component, not in loop.
* Please keep an eye on the JS console during development. This one was generating a non-stop stream of errors as soon as I landed on the user list page. It should likely have been caught during initial development of the feature, and definitely shouldn't have snuck out to production. I don't think it's worth a hotfix to address this, but it's a state I would prefer to avoid, and feels avoidable.

egrojMonroy (approved):

src/admin/modules/user/components/UserList.tsx
egrojMonroy: Got it.

PR #948, opened 2025-02-18
[TM-1677] Implement user verification

Scriptmatico:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Make sure you are requesting to pull a topic/feature/bugfix branch (right side). Don't request your master!
- [ ] Make sure you are making a pull request against the develop branch (left side). Also you should start your branch off our develop.
- [ ] Check the commit's or even all commits' message styles matches our requested structure.
- [ ] Check your code additions will fail neither code linting checks nor unit test.

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?

-
-
-

## Does this introduce a breaking change?

- [ ] Yes
- [ ] No

### DIFF
--- a/src/connections/VerificationUser.ts
+++ b/src/connections/VerificationUser.ts
@@ -1,0 +1,39 @@
+import { createSelector } from "reselect";
+
+import { verifyUser } from "@/generated/v3/userService/userServiceComponents";
+import { verifyUserFetchFailed, verifyUserIsFetching } from "@/generated/v3/userService/userServicePredicates";
+import { ApiDataStore, PendingErrorState } from "@/store/apiSlice";
+import { Connection } from "@/types/connection";
+import { connectionHook } from "@/utils/connectionShortcuts";
+import { selectorCache } from "@/utils/selectorCache";
+
+export const selectVerificationUser = (store: ApiDataStore) => Object.values(store.verifications)?.[0]?.attributes;
+
+type VerificationUserConnection = {
+  isLoading: boolean;
+  requestFailed: PendingErrorState | null;
+  isSuccess: boolean;
+  sendVerifyUser: () => void;
+};
+
+type VerificationUserProps = {
+  token: string;
+};
+
+const verificationUserConnection: Connection<VerificationUserConnection, VerificationUserProps> = {
+  selector: selectorCache(
+    ({ token }) => token,
+    ({ token }) =>
+      createSelector(
+        [verifyUserIsFetching, verifyUserFetchFailed, selectVerificationUser],
+        (isLoading, requestFailed, selector) => ({
+          isLoading,
+          requestFailed,
+          isSuccess: selector?.verified,
+          sendVerifyUser: () => verifyUser({ body: { token } })
+        })
+      )
+  )
+};
+
+export const useVerificationUser = connectionHook(verificationUserConnection);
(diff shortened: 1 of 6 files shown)


src/pages/auth/verify/email/[token].page.tsx
```diff
@@ -19,21 +20,25 @@ const VerifyEmail = () => {
...
 
+  useValueChanged(isSuccess, () => {
+    if (isSuccess) {
+      setVerified(true);
+    }
+  });
+
+  useOnMount(async () => {
+    sendVerifyUser();
+  });
```
Comment (roguenet) [inline src/pages/auth/verify/email/[token].page.tsx:40]: This is superfluous. Connections are meant to be built with automatic loading on mount built in.  See `myUserConnection` in `User.ts` for examples of using `load` and `isLoaded` to achieve this. Remember this about the design of connections: The user of the connection (this component) is declaring "my requirement is a loaded user verification; I don't care how or when it loads, I just need it to render correctly" and the connection's job is to say "I provide a loaded user verification; I'll make sure to get it when I'm loaded and provide it from the cache once it's available". 

The only time that logic in a component should be directly issuing API requests should be based on state changes or user input. E.g. saving data after a user enters something in, deleting a resource based on button push, etc. Simply retrieving the data that is required to fulfill a connection is up to the connection.

Comment (roguenet) [reply src/pages/auth/verify/email/[token].page.tsx:40]: Also a random note about this use of `useOnMount`: 
* There's no reason for that callback to be async.
* When you just want to call a function, you can inline the whole thing: `useOnMount(sendVerifyUser)`. The only time you have be careful about that is if the callback you pass in returns a function, it will be executed when the component unmounts (if the callback returns a function, it passes it as the return value to `useEffect`, as a way to execute cleanup).

src/connections/VerificationUser.ts
```diff
@@ -4,23 +4,28 @@ import { verifyUser } from "@/generated/v3/userService/userServiceComponents";
...
+  isSuccess: boolean | null;
 };
 
 type VerificationUserProps = {
   token: string;
 };
 
 const verificationUserConnection: Connection<VerificationUserConnection, VerificationUserProps> = {
+  load: ({ isSuccess, requestFailed }, { token }) => {
+    console.log(isSuccess, requestFailed);
```
roguenet, line 24: Please don't leave `console.log` usage in code that's ready to merge.

src/connections/VerificationUser.ts
```diff
@@ -4,23 +4,28 @@ import { verifyUser } from "@/generated/v3/userService/userServiceComponents";
...
+    if (isSuccess === null && requestFailed === null) verifyUser({ body: { token } });
```
roguenet, line 25: Please use `==` comparison for null instead of `===`. You want to be able to catch both null and undefined here.

src/pages/auth/verify/email/[token].page.tsx
```diff
@@ -9,34 +9,18 @@ import Confirmation from "@/components/extensive/Confirmation/Confirmation";
...
-    if (isSuccess) {
-      setVerified(true);
-    }
-  });
-
-  useOnMount(async () => {
-    sendVerifyUser();
+  useValueChanged(token, async () => {
+    const { isSuccess } = await loadVerificationUser({ token });
+    setVerified(isSuccess!);
```
Comment (roguenet) [inline src/pages/auth/verify/email/[token].page.tsx:23]: Please don't use connection loaders in components. The hooks are much more efficient. The `useState` for verified is also not needed, and neither is this `useValueChanged` hook. This should be sufficient:
```
const [, { isSuccess: verified }] = useVerificationUser({ token });
```

Comment (roguenet) [reply src/pages/auth/verify/email/[token].page.tsx:23]: The `useValueChanged()` for `fetchFailed` should likely be brought back though.

roguenet (approved):

PR #951, opened 2025-02-19
[TM-1711] add abort signal handling to the notification process

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1711)

### DIFF
--- a/src/admin/components/ResourceTabs/PolygonReviewTab/components/PolygonDrawer/PolygonDrawer.tsx
+++ b/src/admin/components/ResourceTabs/PolygonReviewTab/components/PolygonDrawer/PolygonDrawer.tsx
@@ -25,6 +25,7 @@ import {
 } from "@/generated/apiComponents";
 import { ClippedPolygonResponse, SitePolygon, SitePolygonsDataResponse } from "@/generated/apiSchemas";
 import { parseValidationData } from "@/helpers/polygonValidation";
+import JobsSlice from "@/store/jobsSlice";
 import Log from "@/utils/log";
 
 import AuditLogTable from "../../../AuditLogTab/components/AuditLogTable";
@@ -119,7 +120,12 @@ const PolygonDrawer = ({
     onError: () => {
       setCheckPolygonValidation(false);
       hideLoader();
-      openNotification("error", t("Error! TerraMatch could not review polygons"), t("Please try again later."));
+      if (JobsSlice.currentState.abortJob) {
+        openNotification("warning", t("The operation has been successfully canceled."), t("Canceled"));
+        JobsSlice.reset();
+      } else {
+        openNotification("error", t("Error! TerraMatch could not review polygons"), t("Please try again later."));
+      }
     }
   });
   const mutateSitePolygons = fetchPutV2ENTITYUUIDStatus;
--- a/src/components/elements/Map-mapbox/MapControls/CheckIndividualPolygonControl.tsx
+++ b/src/components/elements/Map-mapbox/MapControls/CheckIndividualPolygonControl.tsx
@@ -12,6 +12,7 @@ import {
 } from "@/generated/apiComponents";
 import { ClippedPolygonResponse, SitePolygonsDataResponse } from "@/generated/apiSchemas";
 import { useValueChanged } from "@/hooks/useValueChanged";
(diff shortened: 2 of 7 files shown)


src/components/elements/Notification/FloatNotification.tsx
```diff
@@ -37,6 +38,20 @@ const FloatNotification = () => {
...
+        uuid: uuid,
+        type: "delayedJobs",
+        attributes: {
+          isAcknowledged: true
+        }
+      }
+    ];
+    JobsSlice.setAbortJob(true);
+    triggerBulkUpdate(job);
+  };
```
roguenet, line 53: This method doesn't seem to have any dependencies within the component, and is pure with regard to the component state. It's usually best to define the function outside of the component body in cases like that. Also, please reserve upper first names for classes - this should be named `abortJobUuid`, or simply `abortJob`.

src/store/jobsSlice.ts
```diff
@@ -52,6 +62,10 @@ export default class JobsSlice {
     this._redux.dispatch(jobsSlice.actions.setJobsProgress({ totalContent, processedContent, progressMessage }));
   }
 
+  static setAbortJob(abortJob: boolean) {
```
roguenet, line 65: I'm not finding any code that calls `setAbortJob(false)` - how does this ever get cleared out?

LimberHope: Thank you for your comment! To reset the state of all values, including abortJob, I’m using the reset action (JobsSlice.reset();), which resets the state to its initial value (INITIAL_STATE). This ensures that when the operation is canceled, the value of abortJob is reset to false and doesn't remain true. This way, we avoid it staying active after the operation has finished. Do you think any adjustments are needed in this approach?

roguenet (dismissed):

src/admin/components/ResourceTabs/PolygonReviewTab/components/PolygonDrawer/PolygonDrawer.tsx
```diff
@@ -119,7 +120,12 @@ const PolygonDrawer = ({
     onError: () => {
       setCheckPolygonValidation(false);
       hideLoader();
-      openNotification("error", t("Error! TerraMatch could not review polygons"), t("Please try again later."));
+      if (JobsSlice.currentState.abortJob) {
```
Scriptmatico, line 123: This code can be optimized for readability and maintainability, 

Could be better this way:

```
const { abortJob } = JobsSlice.currentState;
const notificationType = abortJob ? "warning" : "error";
const notificationMessage = abortJob 
  ? t("The operation has been successfully canceled.") 
  : t("Error! TerraMatch could not review polygons");
const notificationTitle = abortJob ? t("Canceled") : t("Please try again later.");

openNotification(notificationType, notificationMessage, notificationTitle);

if (abortJob) {
  JobsSlice.reset();
}
```

src/components/elements/Map-mapbox/MapControls/CheckIndividualPolygonControl.tsx
```diff
@@ -47,7 +48,12 @@ const CheckIndividualPolygonControl = ({ viewRequestSuport }: { viewRequestSupor
     onError: () => {
       hideLoader();
       setClickedValidation(false);
-      displayNotification(t("Please try again later."), "error", t("Error! TerraMatch could not review polygons"));
+      if (JobsSlice.currentState.abortJob) {
```
Scriptmatico, line 51: We can do the same from PolygonDrawer and refactor this code.

src/components/elements/Map-mapbox/MapControls/CheckPolygonControl.tsx
```diff
@@ -139,7 +139,12 @@ const CheckPolygonControl = (props: CheckSitePolygonProps) => {
       hideLoader();
       setIsLoadingDelayedJob?.(false);
       setClickedValidation(false);
-      displayNotification(t("Please try again later."), "error", t("Error! TerraMatch could not review polygons"));
+      if (JobsSlice.currentState.abortJob) {
```
Scriptmatico, line 142: I think the same applies here

PR #1014, opened 2025-03-07
[TM-1759] research service: add connection

egrojMonroy:


src/connections/SitePolygons.ts
```diff
@@ -0,0 +1,76 @@
...
+import { Connection } from "@/types/connection";
+import { connectionHook, connectionLoader } from "@/utils/connectionShortcuts";
+
+export type SitePolygonConnectionParams = {
+  entityName: string;
+  entityUuid: string;
+};
+
+export type SitePolygonConnection = {
+  sitePolygons?: SitePolygonDto[] | undefined;
```
roguenet, line 23: `| undefined` isn't needed here, the `?` takes care of that.

src/connections/SitePolygons.ts
```diff
@@ -0,0 +1,76 @@
...
+  sitePolygonsHaveFailed: boolean;
+  sitePolygonsAreLoading: boolean;
```
roguenet, line 25: This shouldn't be needed - the base connection system provides this information.

src/connections/SitePolygons.ts
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const sitePolygonQuery = (entityName: string, entityUuid: string) => {
+  const queryParams = {
+    "page[number]": 0, // update once pagination is implemented
```
roguenet, line 30: I'll hopefully be getting this in today.

src/connections/SitePolygons.ts
```diff
@@ -0,0 +1,76 @@
...
+  } as SitePolygonsIndexQueryParams;
+  if (entityName === "projects") {
+    queryParams["projectId[]"] = [entityUuid];
+  } else if (entityName === "sites") {
+    queryParams["siteId[]"] = [entityUuid];
+  }
+  return { queryParams };
+};
+const sitePolygonsSelector = (store: ApiDataStore) =>
+  Object.values(store.sitePolygons || {}).map(resource => resource.attributes);
```
roguenet, line 41: Please use `??` for defaults.

src/connections/SitePolygons.ts
```diff
@@ -0,0 +1,76 @@
...
+  },
+  isLoaded: ({ sitePolygons, sitePolygonsHaveFailed }) => sitePolygons !== null || sitePolygonsHaveFailed,
+  selector: createSelector(
+    [
+      sitePolygonsSelector,
+      sitePolygonsIndexIsFetching(sitePolygonQuery(params.entityName, params.entityUuid)),
+      sitePolygonsIndexFetchFailed(sitePolygonQuery(params.entityName, params.entityUuid))
+    ],
+    (sitePolygons, sitePolygonsAreLoading, sitePolygonsFailed) => ({
+      sitePolygons,
```
roguenet, line 58: This is a little too simple. According to `sitePolygonsSelector` current implementation, this will return all of the site polygons in the store every time. This connection should be filtering out the ones that don't apply to the site / project current being requested for this use of the connection.

src/connections/SitePolygons.ts
```diff
@@ -0,0 +1,76 @@
...
+export const useSitePolygons = (params: SitePolygonConnectionParams) => {
+  const { entityName, entityUuid } = params;
+
+  const memoizedConnection = useMemo(
+    () => sitePolygonsConnection({ entityName, entityUuid }),
+    [entityName, entityUuid]
+  );
+
+  return connectionHook(memoizedConnection)();
+};
```
roguenet, line 73: I don't believe this memoization is needed. Connections are stable and have good caching built in. This should simply be `export const useSitePolygons = connectionHook(sitePolygonsConnection);`

src/connections/SitePolygons.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export const loadSitePolygons = (params: SitePolygonConnectionParams) =>
+  connectionLoader(sitePolygonsConnection(params));
```
roguenet, line 76: This can be simpler as well:
```
export const loadSitePolygons = connectionLoader(sitePolygonsConnection);
```

src/connections/SitePolygons.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const sitePolygonsConnection = (params: SitePolygonConnectionParams): Connection<SitePolygonConnection> => ({
+  load: ({ sitePolygons }) => {
+    if (sitePolygons == null || sitePolygons.length === 0) {
+      const queryParams = sitePolygonQuery(params.entityName, params.entityUuid);
+      sitePolygonsIndex(queryParams);
+    }
+  },
+  isLoaded: ({ sitePolygons, sitePolygonsHaveFailed }) => sitePolygons !== null || sitePolygonsHaveFailed,
+  selector: createSelector(
```
roguenet, line 51: I think the reason that you found a need for the memoization stuff below is because this selector creation is too simple. Please see `selectorCache` and its uses in other connections for the correct pattern of memoizing a selector based on connection params.

src/connections/SitePolygons.ts
egrojMonroy: understand

src/connections/SitePolygons.ts
egrojMonroy: removed

src/connections/SitePolygons.ts
egrojMonroy: We are now sending it, and the param will alway exists.

src/connections/SitePolygons.ts
egrojMonroy: removed

src/connections/SitePolygons.ts
egrojMonroy: changed

src/connections/SitePolygons.ts
egrojMonroy: now selectorcache is implemented with indices

src/connections/SitePolygons.ts
roguenet: Yeah, mapping off the index meta ids does this for you.

src/connections/SitePolygons.ts
```diff
@@ -1,76 +1,84 @@
...
   entityUuid: string;
 };
 
-export type SitePolygonConnection = {
-  sitePolygons?: SitePolygonDto[] | undefined;
-  sitePolygonsHaveFailed: boolean;
-  sitePolygonsAreLoading: boolean;
+export type SitePolygonIndexConnectionProps = SitePolygonConnectionParams & {
+  pageSize?: number;
+  pageNumber?: number;
```
roguenet, line 18: I don't think in this case there's value in having these separate. Only the index props are being used, so these two can be combined.

src/connections/SitePolygons.ts
```diff
@@ -1,76 +1,84 @@
...
-const sitePolygonsConnection = (params: SitePolygonConnectionParams): Connection<SitePolygonConnection> => ({
-  load: ({ sitePolygons }) => {
-    if (sitePolygons == null || sitePolygons.length === 0) {
-      const queryParams = sitePolygonQuery(params.entityName, params.entityUuid);
-      sitePolygonsIndex(queryParams);
+export type SitePolygonIndexConnection<SitePolygonDto> = {
+  sitePolygons?: SitePolygonDto[];
+  indexTotal?: number;
+  fetchFailure?: PendingErrorState | null;
+  refetch: () => void;
```
roguenet, line 30: are indexTotal and refetch useful to the consumers of this connection? There doesn't seem to be any code using them in this PR.

src/connections/SitePolygons.ts
```diff
@@ -1,76 +1,84 @@
...
-  );
+  isLoaded: indexIsLoaded,
+
+  selector: selectorCache(
+    props => sitePolygonCacheKey(props),
+    props =>
+      createSelector(
+        [
+          indexMetaSelector("sitePolygons", sitePolygonQueryParams(props)),
+          (store: ApiDataStore) => store.sitePolygons ?? {},
```
roguenet, line 66: You might have seen this in other places; I probably did it out of habit a couple of times, but this `?? {}` shouldn't be needed. Each of the defined resource stores is created as an empty object when the store is created, so they're guaranteed not to be null.

src/connections/SitePolygons.ts
egrojMonroy: combined

src/connections/SitePolygons.ts
egrojMonroy: ok

roguenet (approved):

PR #1052, opened 2025-03-18
[TM-1857] retry on failed network

egrojMonroy:


cesarLima1 (dismissed):

src/generated/apiFetcher.ts
```diff
@@ -163,11 +163,17 @@ async function loadJob(signal: AbortSignal | undefined, delayedJobId: string, re
     }
 
     return await response.json();
-  } catch (e) {
+  } catch (e: any) {
```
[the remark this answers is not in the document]
egrojMonroy: thanks, thats right. Already changed

src/generated/apiFetcher.ts
```diff
@@ -163,13 +163,21 @@ async function loadJob(signal: AbortSignal | undefined, delayedJobId: string, re
...
+  } catch (e: unknown) {
     Log.error("Delayed Job Fetch error", e);
-    error = {
-      statusCode: response?.status || -1,
-      //@ts-ignore
-      ...(e || {})
-    };
+
+    if (typeof e === "object" && e !== null) {
+      const errorMessage = (e as { message?: string }).message || "";
```
roguenet, line 170: please use `??` for defaults here and on the `isNetworkError` line.

egrojMonroy: Hope this is the last time with that default. 
Changed already for that error Message, but for isNetworkError it evaluates if its true or false. Which the OR would be fine, right?

roguenet (approved):

src/generated/apiFetcher.ts
roguenet: Whoops, my mistake, was reviewing too quickly. What you have now is correct 👍🏻

PR #1074, opened 2025-03-25
Fix/[TM-1852] project organisation filter not working

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1852)
[Task Link](https://gfw.atlassian.net/browse/TM-1870)


roguenet (commented):
Is this really supposed to be adding so much? This looks like it's adding the full project report v3 integration, but doesn't seem to line up with the ticket description.

PR #1101, opened 2025-03-31
[TM-1862] Implement entity delete in v3 for all entities

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-1862)


roguenet (commented):
uses of `useDeleteV2NurseriesUUID` and `useDeleteV2SitesUUID` need to be updated to use v3.

src/components/extensive/Tables/NurseriesTable.tsx
```diff
@@ -57,8 +50,13 @@ const NurseriesTable = ({ project, onFetch, hasAddButton = true }: NurseriesTabl
         primaryButtonProps={{
           children: t("Yes"),
           onClick: () => {
-            deleteNursery({ pathParams: { uuid } });
-            closeModal(ModalId.CONFIRM_NURSERY_DELETION);
+            deleteNursery(uuid)
+              .then(() => {
+                nurseryIndex.refetch();
```
roguenet, line 55: We get some magic here - this refetch is no longer needed; When the delete is recorded in the v3 store, the index connection will note that one of its records is missing and refetch automatically. I would streamline this method like so:
```
onClick: async () => {
  await deleteNursery(uuid);
  closeModal(ModalId.CONFIRM_NURSERY_DELETION);
}
```

src/components/extensive/Tables/SitesTable.tsx
```diff
@@ -56,8 +49,13 @@ const SitesTable = ({ project, hasAddButton = true, onFetch }: SitesTableProps)
         primaryButtonProps={{
           children: t("Yes"),
           onClick: () => {
-            deleteSite({ pathParams: { uuid } });
-            closeModal(ModalId.CONFIRM_SITE_DELETION);
+            deleteSite(uuid)
+              .then(() => {
+                siteIndex.refetch();
```
roguenet, line 54: Same here as with nurseries; the refetch is no longer needed.

src/pages/nursery/[uuid]/components/NurseryHeader.tsx
```diff
@@ -59,8 +49,17 @@ const NurseryHeader = ({ nursery }: NurseryHeaderProps) => {
...
+              .then(() => {
+                router.push("/my-projects");
+                openToast(t("The nursery has been successfully deleted"));
+              })
+              .catch(() => {
+                openToast(t("Something went wrong!"), ToastType.ERROR);
+              })
+              .finally(() => {
+                closeModal(ModalId.CONFIRM_NURSERY_DELETION);
+              });
```
Comment (roguenet) [inline src/pages/nursery/[uuid]/components/NurseryHeader.tsx:62]: :art: There's no reason we can't use async functions for onClick callbacks:
```
onClick: async () => {
  try {
    await deleteNursery(nursery?.uuid);
    openToast(t("The nursery has been successfully deleted"));
  } catch () {
    openToast(t("Something went wrong!"), ToastType.ERROR);
  }
  closeModal(ModalId.CONFIRM_NURSERY_DELETION);
}
```

src/components/extensive/Tables/SitesTable.tsx
```diff
@@ -56,8 +49,13 @@ const SitesTable = ({ project, hasAddButton = true, onFetch }: SitesTableProps)
         primaryButtonProps={{
           children: t("Yes"),
           onClick: () => {
-            deleteSite({ pathParams: { uuid } });
-            closeModal(ModalId.CONFIRM_SITE_DELETION);
+            deleteSite(uuid)
```
Scriptmatico, line 52: Why is this function using promises instead of async/await? Is really needed?

src/connections/Entity.ts
```diff
@@ -263,6 +273,11 @@ export const useFullProjectReport = connectionHook(fullProjectReportConnection);
 const lightProjectReportConnection = createGetEntityConnection<ProjectReportLightDto>("projectReports", false);
 export const loadLightProjectReport = connectionLoader(lightProjectReportConnection);
 export const useLightProjectReport = connectionHook(lightProjectReportConnection);
+export const deleteProjectReport = connectedResourceDeleter(
```
Scriptmatico, line 276: "projectReports" is being used in too many lines, we should consider declare a const and stop using magic strings.

src/components/extensive/Tables/SitesTable.tsx
Scriptmatico: same has a comment bellow from Nathan you can use a similar code on this secction instead of callbacks:

`onClick: async () => {
  try {
    await deleteSite(nursery?.uuid);
    openToast(t("The site has been successfully deleted"));
  } catch () {
    openToast(t("Something went wrong!"), ToastType.ERROR);
  }
  closeModal(ModalId.CONFIRM_SITE_DELETION);
}`

src/connections/Entity.ts
roguenet: Normally I agree with this sort of feedback. In this case I think it's actually totally fine because the string isn't all that magical: The function being called in these cases is well typed, and only accepts a certain set of values:
image

roguenet (approved):

roguenet (changes requested):
Please update the SiteHeader as well

roguenet (approved):

PR #1142, opened 2025-04-11
[TM-1774] remove use of migrated flag in fe

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1774


src/pages/entity/[entityName]/edit/[uuid]/EditEntityForm.tsx
```diff
@@ -64,7 +64,7 @@ const EditEntityForm = ({ entityName, entityUUID, entity, formData }: EditEntity
     () => defaults(formData?.update_request?.content ?? {}, formData?.answers),
     [formData?.answers, formData?.update_request?.content]
   );
-  const defaultValues = useNormalizedFormDefaultValue(sourceData, formSteps, entity.migrated);
```
Comment (roguenet) [inline src/pages/entity/[entityName]/edit/[uuid]/EditEntityForm.tsx:67]: Did you investigate how this hook is using the flag? Making sure that simply removing this doesn't break display of old entities was the main point of this ticket, and I want to make sure you dug into that and checked out some old models.

Comment (Scriptmatico) [reply src/pages/entity/[entityName]/edit/[uuid]/EditEntityForm.tsx:67]: The investigation was done, that property is only used in a function for the EntityForm.ts forms, and it is only used for this case within customForm.ts

```
case FieldType.Conditional: {
  if (isMigrated && typeof values[[field.name](http://field.name/)] !== "boolean") values[[field.name](http://field.name/)] = true;
  field?.fieldProps.fields.map(f => normalizedFieldDefaultValue(values, f, isMigrated));
  break;
}
```

`export function normalizedFormDefaultValue<T = any>(values?: T, steps?: FormStepSchema[], isMigrated?: boolean): T {`

The param isMigrated is opcional

Comment (roguenet) [reply src/pages/entity/[entityName]/edit/[uuid]/EditEntityForm.tsx:67]: Wild, ok. Well, let's also remove the `isMigrated` optional param from that method, as that should now be dead code.

PR #1198, opened 2025-05-06
[TM-1995] Migrate project pitches to V3

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1995


src/admin/apiProvider/dataProviders/pitchDataProviders.ts
```diff
@@ -1,67 +1,43 @@
...
-      const response = await fetchGetV2ProjectPitchesUUID({
-        //@ts-ignore
-        pathParams: { uuid: params.id }
-      });
-      //@ts-ignore
-      return { data: { ...response.data, id: response.data.uuid } };
-    } catch (err) {
-      throw getFormattedErrorForRA(err as GetV2ProjectPitchesUUIDError);
+    const { requestFailed, projectPitch } = await loadProjectPitch({ uuid: params.id });
+    console.log("getOne response", projectPitch, requestFailed);
```
roguenet, line 34: Please don't leave debug logging in - there should be no uses of `console.log` in the PR.

src/admin/apiProvider/utils/listing.ts
```diff
@@ -103,3 +105,8 @@ export const apiListResponseToRAListResult = (response: ApiListResponse): Extend
     included: response?.included
   };
 };
+
+export const projectPitchesListResult = ({ data, indexTotal }: ProjectsPitchesConnection) => ({
+  data: data?.map((pitch: ProjectPitchDto) => ({ ...pitch, id: pitch.uuid })) as any,
+  total: indexTotal ?? 0
+});
```
roguenet, line 112: I'd argue that this  method isn't really needed, but even if you do want to keep it separate please move it to to the pitchDataProvider - it's not a general utility that will be needed in other places.

src/admin/modules/pitch/components/PitchesList.tsx
```diff
@@ -64,20 +64,20 @@ const ApplicationDataGrid = () => {
...
       <SimpleChipFieldArray
-        source="restoration_intervention_types"
+        source="restorationInterventionTypes"
         label="Restoration Intervention Types"
         choices={optionToChoices(getRestorationInterventionTypeOptions())}
       />
-      <SelectField source="project_country" label="Countries" choices={countryChoices} />
-      <DateField source="created_at" label="Date Added" locales="en-GB" />
+      <SelectField source="projectCountry" label="Countries" choices={countryChoices} />
+      <DateField source="createdAt" label="Date Added" locales="en-GB" />
```
roguenet, line 80: Please double check that these fields are all included. For instances, `createdAt` does not appear to be on the v3 DTO, unless I'm missing something.

src/connections/Entity.ts
```diff
@@ -91,7 +91,7 @@ export type EntityIndexConnection<T extends EntityDtoType> = {
   included?: JsonApiResource[];
 };
 
-type EntityIndexFilterKey = keyof Omit<
+export type EntityIndexFilterKey = keyof Omit<
```
roguenet, line 94: As mentioned in the v3 PR, I'd like to avoid re-using the entity query params for the project pitches controller. It should get its own that is specific to pitches.

src/connections/ProjectPitch.ts
```diff
@@ -0,0 +1,132 @@
...
+import { selectorCache } from "@/utils/selectorCache";
+
+const projectPitchesSelector = () => (store: ApiDataStore) =>
+  store["projectPitches"] as StoreResourceMap<ProjectPitchDto>;
+
+type ProjectPitchConnection = {
+  isLoading: boolean;
+  requestFailed: PendingErrorState | null;
+  isSuccess: boolean;
+  projectPitch: ProjectPitchDto | any;
```
roguenet, line 29: Why is this `any` here? We should be able to avoid uses of `any` in new v3 code.

src/connections/ProjectPitch.ts
```diff
@@ -0,0 +1,132 @@
...
+};
+
+type ProjectPitchConnectionProps = {
+  uuid: string;
+};
+
+const projectPitchConnection: Connection<ProjectPitchConnection, ProjectPitchConnectionProps> = {
+  load: ({ projectPitch }, { uuid }) => {
+    if (!projectPitch) projectPitchesGetUUIDIndex({ pathParams: { uuid } });
```
roguenet, line 38: Two things here:
* Please remember to use explicit null checks: `projectPitch == null`
* This is not sufficient. If the fetch fails, the request is complete but the pitch will be null so this will just continuously retry to failing fetch. See other connections for examples. In most cases, extracting the `isLoaded` line below to a separate function and calling it here with the connection state is enough.

src/connections/ProjectPitch.ts
```diff
@@ -0,0 +1,132 @@
...
+  projectPitchesGetUUIDIndexFetchFailed,
+  projectPitchesGetUUIDIndexIsFetching
+} from "@/generated/v3/entityService/entityServiceSelectors";
+import { getStableQuery } from "@/generated/v3/utils";
+import ApiSlice, { ApiDataStore, PendingErrorState, StoreResourceMap } from "@/store/apiSlice";
+import { Connection } from "@/types/connection";
+import { connectionLoader } from "@/utils/connectionShortcuts";
+import { selectorCache } from "@/utils/selectorCache";
+
+const projectPitchesSelector = () => (store: ApiDataStore) =>
```
roguenet, line 22: I don't see any reason to have this intermediate function. This selector can be
```
const projectPitchesSelector = ({ projectPitches }: ApiDataStore) => projectPitches;
```
The in the selector below, you'd use `projectPitchesSelector` instead of `projectPitchesSelector()`

src/connections/ProjectPitch.ts
```diff
@@ -0,0 +1,132 @@
...
+  store["projectPitches"] as StoreResourceMap<ProjectPitchDto>;
```
roguenet, line 23: This `as` typing should not be needed - the store is very thoroughly typed.

src/connections/ProjectPitch.ts
```diff
@@ -0,0 +1,132 @@
...
+      createSelector(
+        [
+          projectPitchesGetUUIDIndexIsFetching({ pathParams: { uuid } }),
+          projectPitchesGetUUIDIndexFetchFailed({ pathParams: { uuid } }),
+          projectPitchesSelector()
+        ],
+        (isLoading, requestFailed, selector) => ({
+          isLoading,
+          requestFailed,
+          isSuccess: selector?.organisationId != null,
```
roguenet, line 54: I'm not sure I see a need for this `isSuccess` boolean. The presence of the project pitch should be enough to indicate success. It also doesn't look like it'd work, since `selector` is a store of project pitches, not an individual pitch.

src/connections/ProjectPitch.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export type ProjectPitchIndexConnectionProps = {
+  pageSize?: number;
+  pageNumber?: number;
+  sortField?: string;
+  sortDirection?: "ASC" | "DESC";
+  filter?: Partial<Record<EntityIndexFilterKey, string>>;
+};
+
+const entityIndexQuery = (props?: ProjectPitchIndexConnectionProps) => {
```
roguenet, line 76: This should be `projectPitchIndexQuery` or something similar. It's a bit confusing, but in these contexts, "entity" specifically means projects / sites / nurseries / reports. ProjectPitch doesn't count as an "Entity"

src/connections/ProjectPitch.ts
```diff
@@ -35,24 +32,23 @@ type ProjectPitchConnectionProps = {
 
 const projectPitchConnection: Connection<ProjectPitchConnection, ProjectPitchConnectionProps> = {
   load: ({ projectPitch }, { uuid }) => {
-    if (!projectPitch) projectPitchesGetUUIDIndex({ pathParams: { uuid } });
+    if (projectPitch == null) projectPitchGet({ pathParams: { uuid } });
   },
 
-  isLoaded: ({ projectPitch }) => projectPitch !== undefined,
+  isLoaded: ({ projectPitch }) => projectPitch !== null,
```
roguenet, line 38: [As mentioned in my previous review](https://github.com/wri/wri-terramatch-website/pull/1198#discussion_r2085168841), this isn't good enough. Two things:
* Please always use double equals for null comparison. It's always safer to catch both null and undefined since it's so easy for one to sneak in where you expect the other (`!= null` instead of `!== null`). 
* If the request fails, this code will cause the connection continuous try to re-fetch the resource. Please see other connections for an example. The `isLoaded` method pretty much always both checks for existence of the resource, and the fetch failure at a minimum, and the same things need to be checked before calling the API request in the `load()` function.

src/admin/apiProvider/dataProviders/pitchDataProviders.ts
```diff
@@ -1,67 +1,48 @@
...
-  "restoration_intervention_types",
-  "total_hectares",
-  "total_trees",
-  "capacity_building_needs",
-  "created_at",
-  "updated_at",
-  "deleted_at"
-];
+const projectPitchesListResult = ({ data, indexTotal }: ProjectsPitchesConnection) => ({
+  data: data?.map((pitch: ProjectPitchDto) => ({ ...pitch, id: pitch.uuid })) as any,
```
roguenet, line 25: I don't think `as any` is needed here; we don't seem to be using it in the other data providers.

src/admin/apiProvider/dataProviders/pitchDataProviders.ts
```diff
@@ -1,67 +1,48 @@
...
-        //@ts-ignore
-        pathParams: { uuid: params.id }
-      });
-      //@ts-ignore
-      return { data: { ...response.data, id: response.data.uuid } };
-    } catch (err) {
-      throw getFormattedErrorForRA(err as GetV2ProjectPitchesUUIDError);
+    const { requestFailed, projectPitch } = await loadProjectPitch({ uuid: params.id });
+    if (requestFailed != null) {
+      throw v3ErrorForRA("Nursery get fetch failed", requestFailed);
```
roguenet, line 42: This should be `Project Pitch`, not `Nursery`

src/admin/modules/pitch/components/PitchEdit.tsx
```diff
@@ -26,33 +26,29 @@ const PitchEdit = () => {
...
-          source="capacity_building_needs"
+          source="capacityBuildingNeeds"
           choices={optionToChoices(getCapacityBuildingNeedOptions())}
           label="Capacity Building Needs"
           fullWidth
         />
-        <NumberInput source="total_hectares" label="Total Hectares to be restored" fullWidth />
-        <NumberInput source="total_trees" label="Total number of trees to be grown" fullWidth />
+        <NumberInput source="totalHectares" label="Total Hectares to be restored" fullWidth />
+        <NumberInput source="totalTrees" label="Total number of trees to be grown" fullWidth />
```
roguenet, line 51: I just realized we have a problem here that doesn't exist for the other entities. Most entities only get updated through the form system, but via this Admin control, pitches can be edited directly. I think the solution for this PR is to go to the `pitchDataProvider` and change the `update` method to move these camel case properties back to snake case for the send to v3.

src/connections/ProjectPitch.ts
```diff
@@ -0,0 +1,133 @@
...
+    props =>
+      createSelector(
+        [
+          projectPitchIndexIndexMeta("projectPitches", projectPitchesIndexParams(props)),
+          projectPitchIndexFetchFailed(projectPitchesIndexParams(props)),
+          projectPitchesSelector
+        ],
+        (indexMeta, fetchFailure, selector) => {
+          console.log(indexMeta);
+          const refetch = () => ApiSlice.pruneIndex("projectPitches", "");
```
roguenet, line 117: Is anything using this `refetch` method? I'm not finding any users in this PR, and I can't think of why it would be needed for this connection.

roguenet (commented):
One last thing: you've removed a lot of APIs from PHP, but I don't see that reflected here. Please run `yarn generate:api` locally with the correct PHP branch loaded, and commit the changes.

src/generated/v3/entityService/entityServiceComponents.ts
```diff
@@ -105,28 +95,84 @@ export type ProjectPitchIndexResponse = {
...
+    ProjectPitchesIndexError,
     undefined,
     {},
-    ProjectPitchIndexQueryParams,
+    ProjectPitchesIndexQueryParams,
     {}
   >({ url: "/entities/v3/projectPitches", method: "get", ...variables, signal });
 
-export type ProjectPitchGetPathParams = {
+export type AdminProjectPitchesIndexQueryParams = {
```
roguenet, line 112: I didn't expect to see this - didn't we get rid of `admin` routes? I'm also surprised to see `pageNumber` instead of `page[number]` in these. Is your current v3 branch up to date with your PR and running the latest code?

src/generated/v3/entityService/entityServiceComponents.ts
```diff
@@ -155,7 +201,7 @@ export type ProjectPitchGetError = Fetcher.ErrorWrapper<
     }
 >;
 
-export type ProjectPitchGetResponse = {
+export type ProjectPitchesGetUUIDIndexResponse = {
```
roguenet, line 204: These names are also weird. Maybe your operationIds are not up to date in v3?

roguenet (dismissed):

roguenet (dismissed):

roguenet (approved):

PR #1199, opened 2025-05-07
[TM-2009] if no uuid provided then no-op

egrojMonroy:
This adds to connection a no-op case when there is no uuid provided. 
the useFullEntity previously if no uuid provided, returns multiple entities ( not sure if all ).

### DIFF
--- a/src/connections/Entity.ts
+++ b/src/connections/Entity.ts
@@ -156,6 +156,9 @@ const createGetEntityConnection = <T extends EntityDtoType, U extends EntityUpda
   entityName: U["type"]
 ): Connection<EntityConnection<T, U>, EntityConnectionProps> => ({
   load: (connection, props) => {
+    if (!props.uuid || props.uuid.trim() === "") {
+      return;
+    }
     if (!entityIsLoaded(true)(connection, props)) entityGet(specificEntityParams(entityName, props.uuid));
   },
 


src/connections/Entity.ts
```diff
@@ -156,6 +156,9 @@ const createGetEntityConnection = <T extends EntityDtoType, U extends EntityUpda
   entityName: U["type"]
 ): Connection<EntityConnection<T, U>, EntityConnectionProps> => ({
   load: (connection, props) => {
+    if (!props.uuid || props.uuid.trim() === "") {
+      return;
+    }
```
roguenet, line 161: The more typical way to do this is in the `isLoaded` predicate. If the requirements to load a connection aren't provided in the props, it's considered to be loaded until those requirements are met. It looks like that's already implemented for the null case, so maybe this is needed
```
diff --git a/src/connections/Entity.ts b/src/connections/Entity.ts
index 93f32146..8d8d5f9a 100644
--- a/src/connections/Entity.ts
+++ b/src/connections/Entity.ts
@@ -147,7 +147,7 @@ const entityIsLoaded =
     { entity, entityIsDeleted, fetchFailure }: EntityConnection<T, U>,
     { uuid }: EntityConnectionProps
   ) => {
-    if (uuid == null || entityIsDeleted || fetchFailure != null) return true;
+    if (uuid || uuid.trim() === "" || entityIsDeleted || fetchFailure != null) return true;
     if (entity == null) return false;
     return !requireFullEntity || !entity.lightResource;
   };
```

When `isLoaded` return true, `load()` isn't called, so that should fix this up.

egrojMonroy: haven't contemplate that segment. Removed previous implementation and added this one. Is working good.

roguenet (approved):

PR #1215, opened 2025-05-14
[TM-2048] Application form fix

cesarLima1:
https://gfw.atlassian.net/browse/TM-2048

### DIFF
--- a/src/components/extensive/WizardForm/utils.ts
+++ b/src/components/extensive/WizardForm/utils.ts
@@ -46,10 +46,19 @@ export const getSchemaFields = (fields: FormField[]) => {
         }
       });
     } else {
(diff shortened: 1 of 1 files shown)


egrojMonroy (approved):

src/components/extensive/WizardForm/utils.ts
```diff
@@ -46,10 +46,19 @@ export const getSchemaFields = (fields: FormField[]) => {
         }
       });
     } else {
-      schema[field.name] = field.validation?.nullable().label(" ");
+      if (!field.validation) {
```
roguenet, line 49: I won't require you change this after the fact, but please remember to use explicit null checks: `if (field.validation == null)`

PR #1232, opened 2025-05-22
[TM-2016] Add filtering based on nothing to report

Scriptmatico:
https://gfw.atlassian.net/browse/TM-2016

### DIFF
--- a/src/constants/options/status.ts
+++ b/src/constants/options/status.ts
@@ -93,11 +93,11 @@ export const getChangeRequestStatusOptions = (t: typeof useT | Function = (t: st
 export const getNothingReportOptions = (t: typeof useT | Function = (t: string) => t) =>
   [
     {
-      value: "reportInfo",
+      value: "1",
       title: t("Reported Info")
     },
     {
-      value: "nothingToReport",
+      value: "-1",
       title: t("Nothing to Report")
     }
   ] as Option[];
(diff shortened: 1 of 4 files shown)


src/constants/options/status.ts
```diff
@@ -89,3 +89,15 @@ export const getChangeRequestStatusOptions = (t: typeof useT | Function = (t: st
...
   ] as Option[];
+
+export const getNothingReportOptions = (t: typeof useT | Function = (t: string) => t) =>
+  [
+    {
+      value: "reportInfo",
+      title: t("Reported Info")
+    },
+    {
+      value: "nothingToReport",
```
roguenet, line 100: Oh, this helps explain some of what's going on in the v3 controller, but why use these arbitrary strings instead of a simple boolean?

src/constants/options/status.ts
```diff
@@ -89,3 +89,15 @@ export const getChangeRequestStatusOptions = (t: typeof useT | Function = (t: st
...
+export const getNothingReportOptions = (t: typeof useT | Function = (t: string) => t) =>
+  [
+    {
+      value: "1",
+      title: t("Reported Info")
+    },
+    {
+      value: "-1",
+      title: t("Nothing to Report")
+    }
```
roguenet, line 102: Shouldn't these be `true` / `false` (or if strings are required, `"true"`, `"false"`)?

roguenet: I'll pull down your branch and see if I can figure it out either this afternoon or tomorrow morning. Please in the future, if you're needing to spend hours trying to track something like this down, pull me in. That definitely is a case of spending too long trying to figure something out without asking for help.

PR #1244, opened 2025-05-28
[TM-2035] Add organization type parameter to tree species data hook

cesarLima1:
https://gfw.atlassian.net/browse/TM-2035

### DIFF
--- a/src/pages/dashboard/hooks/useDashboardData.ts
+++ b/src/pages/dashboard/hooks/useDashboardData.ts
@@ -287,7 +287,7 @@ export const useDashboardData = (filters: any) => {
   }, [filters.uuid, projectEmploymentData, jobsCreatedData]);
 
   const { treeSpeciesData: projectTreeSpeciesData, isLoading: isLoadingProjectTreeSpecies } =
-    useDashboardTreeSpeciesData(filters.uuid, projectFullDto?.treesGrownGoal);
+    useDashboardTreeSpeciesData(filters.uuid, projectFullDto?.treesGrownGoal, projectFullDto?.organisationType);
 
   const combinedHectaresData = useMemo(() => {
     if (filters.uuid && projectTreeSpeciesData) {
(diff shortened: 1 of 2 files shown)


src/pages/dashboard/hooks/useDashboardTreeSpeciesData.ts
```diff
@@ -57,8 +58,7 @@ export const useDashboardTreeSpeciesData = (
     entities.forEach(({ uuid, dueAt }) => {
       reportsByUuid.set(uuid, {
         dueAt: dueAt,
-        // TODO: Now that this is correctly typed, we can see that organisationType is not on this DTO yet.
-        organisationType: "non-profit-organization"
+        organisationType: organisationType
```
roguenet, line 61: Since it's the same for all site reports, I don't think it needs to be mapped here, but can be just directly referenced in effect hook below.

src/pages/dashboard/hooks/useDashboardData.ts
```diff
@@ -287,7 +287,7 @@ export const useDashboardData = (filters: any) => {
   }, [filters.uuid, projectEmploymentData, jobsCreatedData]);
 
   const { treeSpeciesData: projectTreeSpeciesData, isLoading: isLoadingProjectTreeSpecies } =
-    useDashboardTreeSpeciesData(filters.uuid, projectFullDto?.treesGrownGoal);
+    useDashboardTreeSpeciesData(filters.uuid, projectFullDto?.treesGrownGoal, projectFullDto?.organisationType);
```
roguenet, line 290: Should this still default to `non-profit-organization`?

src/pages/dashboard/hooks/useDashboardTreeSpeciesData.ts
cesarLima1: It's not mapped anymore

PR #1264, opened 2025-06-04
[TM-2083] change tree restoration v3

egrojMonroy:


src/connections/DashboardTreeRestorationGoal.ts
```diff
@@ -0,0 +1,63 @@
...
+  load: (connection, props) => {
+    if (!treeRestorationGoalIsLoaded(connection)) {
+      getTreeRestorationGoal({ queryParams: props });
+    }
+  },
+
+  isLoaded: treeRestorationGoalIsLoaded,
+
+  selector: selectorCache(
+    props => indexCacheKey(props)?.replace(/%5B%5D/g, ""),
```
roguenet, line 40: This shouldn't need `?` - `getStableQuery()` always returns a string.

src/connections/DashboardTreeRestorationGoal.ts
```diff
@@ -0,0 +1,63 @@
...
+          (store: ApiDataStore) =>
+            store.treeRestorationGoals?.[indexCacheKey(props)?.replace(/%5B%5D/g, "")]?.attributes,
+          getTreeRestorationGoalFetchFailed({ queryParams: props })
+        ],
+        (indexMeta, fetchFailure) => {
+          const refetch = () => ApiSlice.pruneIndex("treeRestorationGoals", "");
+          if (indexMeta == null) return { refetch, fetchFailure };
+
+          return {
+            data: indexMeta as TreeRestorationGoalDto,
```
roguenet, line 53: I see lots of `index` naming spread around here, likely copied from another connection. This endpoint isn't an index, and this data isn't an index meta, so let's clean up those names.

src/connections/DashboardTreeRestorationGoal.ts
```diff
@@ -0,0 +1,63 @@
...
+            fetchFailure,
+            refetch
+          };
+        }
+      )
+  )
+};
+
+export const loadTreeRestorationGoal = connectionLoader(treeRestorationGoalConnection);
```
roguenet, line 62: This loader isn't used; please remove it.

src/constants/dashboardConsts.ts
```diff
@@ -71,37 +71,37 @@ export const DUMMY_DATA_FOR_CHART_MULTI_LINE_CHART = [
...
-      { time: "2023-07-01T00:00:00.000Z", value: 10100000, name: "Non Profit" },
-      { time: "2024-02-01T00:00:00.000Z", value: 15500000, name: "Non Profit" },
-      { time: "2024-07-01T00:00:00.000Z", value: 20300000, name: "Non Profit" }
+      { time: "2022-09-01T00:00:00.000Z", value: 0, name: "Non Profit" },
+      { time: "2022-10-01T00:00:00.000Z", value: 0, name: "Non Profit" },
+      { time: "2022-12-01T00:00:00.000Z", value: 0, name: "Non Profit" },
+      { time: "2023-01-01T00:00:00.000Z", value: 0, name: "Non Profit" },
+      { time: "2023-07-01T00:00:00.000Z", value: 0, name: "Non Profit" },
+      { time: "2024-02-01T00:00:00.000Z", value: 0, name: "Non Profit" },
+      { time: "2024-07-01T00:00:00.000Z", value: 0, name: "Non Profit" }
```
roguenet, line 104: I'm curious why these mock values were all changed? Also, are they still needed? If no longer needed, please remove them.

src/pages/dashboard/components/HeaderDashboard.tsx
```diff
@@ -350,13 +349,13 @@ const HeaderDashboard = (props: HeaderDashboardProps) => {
...
         </Text>
-        <If condition={!isHomepage}>
+        <When condition={!isHomepage}>
           <Text variant="text-16" className="absolute top-3 right-3 text-white">
             {t("Last Updated on {date}", {
               date: isValidDate(lastUpdatedAt) ? new Date(String(lastUpdatedAt))?.toISOString().split("T")[0] : ""
             })}
           </Text>
-        </If>
+        </When>
```
roguenet, line 358: :art: in general, I'd like to get rid of this react-if library. It seems to just be syntactic sugar with some annoying performance side effects. Not a big deal to leave this alone, but FYI I've been removing it as I refactor components. This can a simple ternary instead, and then react doesn't have to pretend it's going to render the Text component when it's not:
```
{ 
  isHomepage ? null : <Text ...
  </Text>
}
```

src/utils/dashboardUtils.ts
```diff
@@ -164,12 +164,13 @@ export const createQueryParams = (filters: any) => {
...
 
-export const getRestorationGoalResumeData = (data: DashboardTreeRestorationGoalResponse) => {
-  return [
+export const getRestorationGoalResumeData = (data: TreeRestorationGoalDto) => {
+  const result = [
     { name: "Total", value: data?.totalTreesGrownGoal, color: "#13487A" },
     { name: "Enterprise", value: data?.forProfitTreeCount, color: "#7BBD31" },
     { name: "Non Profit", value: data?.nonProfitTreeCount, color: "#B9EDFF" }
   ];
+  return result;
```
roguenet, line 173: No reason to assign to a variable just to turn around and return it (my IDE would even give a soft warning on this line because of it). I'd either go back to the `return [` statement, or streamline this even more and remove the return:
```
export const getRestorationGoalResumeData = (data: TreeRestorationGoalDto) => [
  ...
];
```

roguenet (approved):

roguenet (changes requested):
Whoops, didn't mean to approve

src/connections/DashboardTreeRestorationGoal.ts
egrojMonroy: renamed

src/connections/DashboardTreeRestorationGoal.ts
egrojMonroy: removed

roguenet (approved):

PR #1325, opened 2025-07-02
[TM-2134] Connection Factory

roguenet:
https://gfw.atlassian.net/browse/TM-2134

As usual, please note that the build will fail until the v3 PR has been merged and deployed.

As we've talked about a few times over the last couple of weeks, this is the big connection factory project. A few things to look at in particular:
* [Codegen](https://github.com/wri/wri-terramatch-website/pull/1325/files#diff-6a0066a86e3fd9946ed6d76d47a577b353fb9bd6ef0c8140b0a457574ada4a83) for v3 has had an update. Instead of generating a `Selectors.ts` file for each service, the components file is getting rewritten to generate an instance of the new [`V3ApiEndpoint`](https://github.com/wri/wri-terramatch-website/pull/1325/files#diff-87316b44e96a55288380ede539516075b021723e15a55c04048f8f3bd30394f4R94) class, which contains useful info like the URL and method, and provides selectors. This object is then used directly in hand-made connections, or passed into the factory as needed. This greatly simplified creating connections via factory because we don't have to pass in all the selectors and the fetcher individually, and some types that I was needing to supply explicitly can be more easily inferred.
* A [typical connection](https://github.com/wri/wri-terramatch-website/pull/1325/files#diff-7f6c3a3b8dc3dcf01e834fb0bfcbc58b48f6fc1ee1e6de8320ff7b27479b70ab) has gotten vastly simpler. This will make connections have a more predictable shape, reduce errors in connection creation and just overall make life easier for integrating V3 endpoints. 
* The connection factory handles a lot of typical usecases. It's started with a call to the exported [v3Resource](https://github.com/wri/wri-terramatch-website/pull/1325/files#diff-9cf1de3c2ec45fb4c5899d1d68804733b2a5abdb6474620cb23eca45ef8f9d06R233) method, which exposes methods for the type of connection you want to create. It returns an instance of [the factory class](https://github.com/wri/wri-terramatch-website/pull/1325/files#diff-9cf1de3c2ec45fb4c5899d1d68804733b2a5abdb6474620cb23eca45ef8f9d06R300) to chain in features that you need for this particular connection. Finally, a call to `buildConnection()` creates the final connection. Note that this method takes an optional string value that you can pass in to enable debug logging (with that string in the logs for easy filtering in the browser console). The methods that are useful and accessible for factory chaining start with [isLoading](https://github.com/wri/wri-terramatch-website/pull/1325/files#diff-9cf1de3c2ec45fb4c5899d1d68804733b2a5abdb6474620cb23eca45ef8f9d06R492). The factory methods all modify the props of the connection, the returned shape of the connection or both.
* The `create` case is handled with its own connection type, created via factory. See examples in [login](https://github.com/wri/wri-terramatch-website/pull/1325/files#diff-bcf35fb9a7b626640c1a4d8646f7625248a31463d430219c358c84cf08c9ff1c) or [signup](https://github.com/wri/wri-terramatch-website/pull/1325/files#diff-4c8e0254ae55bb897e018b635c9138bfb7eca599f3bc36715b25107eb03665ab)


egrojMonroy (dismissed):

LimberHope (dismissed):

cesarLima1 (dismissed):

src/components/generic/Navbar/NavbarContent.tsx
```diff
@@ -67,19 +66,14 @@ const NavbarContent = ({ handleClose, ...rest }: NavbarContentProps) => {
...
-          <NavbarItem href="/auth/login" iconName={IconNames.LOGIN} onClick={handleClose}>
-            {t("Sign in")}
-          </NavbarItem>
-        </Then>
-      </If>
-      <If condition={isLoggedIn}>
-        <MyAccountDropdown isLoggedIn={isLoggedIn} />
-      </If>
+      {navItems.length === 0 ? undefined : <div className="hidden h-4 w-[1px] bg-neutral-500 sm:mx-2 sm:block" />}
+      {login == null ? (
```
Scriptmatico, line 70: Login cannot be undefined here?

roguenet: It can be undefined. I'm not sure I understand your question. `== null` captures both undefined and null and is generally used as a safe way to check for both.

src/components/extensive/Tables/CompletedReportsTable.tsx
```diff
@@ -147,9 +93,31 @@ const CompletedReportsTable = ({ modelName, modelUUID, onFetch }: CompletedRepor
             </Button>
           )
         }
-      ]}
-    />
+      ] as ColumnDef<SiteReportLightDto | NurseryReportLightDto>[],
+    [format, modelName, t]
   );
+
+  if (modelName === "sites") {
```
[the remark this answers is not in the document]
roguenet: If we ended up with more than two types that this component supports, I would move a structure like this one. Since this component is strictly typed to `"sites" | "nurseries"`, I think the way it is now is more readable for now.

PR #1344, opened 2025-07-15
[TM-2125] Migration hectares under restoration using v3 endpoint add validation

Scriptmatico:


src/pages/dashboard/hooks/useDashboardData.ts
```diff
@@ -697,7 +695,7 @@ export const useDashboardData = (filters: any) => {
     hectaresUnderRestoration: finalHectaresUnderRestoration,
     isLoadingJobsCreated: isLoadingJobsCreated || (filters.uuid && isLoadingProjectEmployment),
     isLoadingTreeRestorationGoal: treeRestorationGoalLoaded ?? (filters.uuid && isLoadingProjectTreeSpecies),
-    isLoadingX: isDashboardHectareRestorationLoaded ?? (filters.uuid && isLoadingProjectHectares),
+    isLoadingX: filters.uuid && isLoadingProjectHectares,
```
roguenet, line 698: I don't think there's a good reason to avoid using the loaded flag from the hectares restoration connection. It's easy to pass back from the `useHectareRestoration` updated hook, and it will work as expected; if the filter does not have valid params, the connection will show that it's loaded immediately without doing a round trip to the server. That's how the enabled flag works. In the case where real data is being loaded, it's still useful to pass that state down here.

roguenet (approved):

PR #1360, opened 2025-07-24
[TM-2295] project view dashboard using v3 service

cesarLima1:
https://gfw.atlassian.net/browse/TM-2295

### DIFF
--- a/src/connections/DashboardEntity.ts
+++ b/src/connections/DashboardEntity.ts
@@ -2,22 +2,16 @@ import { v3Resource } from "@/connections/util/apiConnectionFactory";
 import { connectionHook } from "@/connections/util/connectionShortcuts";
 import {
   dashboardEntityIndex,
-  DashboardEntityIndexQueryParams,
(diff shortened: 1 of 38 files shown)


src/connections/DashboardEntity.ts
```diff
@@ -15,4 +19,22 @@ const dashboardProjectsConnection = v3Resource("dashboardProjects", dashboardEnt
   .refetch(() => ApiSlice.pruneIndex("dashboardProjects", ""))
   .buildConnection();
 
+const createDashboardProjectGetConnection = <D extends DashboardProjectsFullDto | DashboardProjectsLightDto>(
+  requireFullEntity: boolean = true
```
roguenet, line 23: Is this just here for future work? It doesn't seem to be used in this PR.

cesarLima1: I was following the logic used in similar examples, but I think this is a better approach.

roguenet (approved):

PR #1385, opened 2025-08-04
[TM-2352] add visual pagination for site polygons list in admin and pd

cesarLima1:
https://gfw.atlassian.net/browse/TM-2352

### DIFF
--- a/src/admin/components/ResourceTabs/PolygonReviewTab/components/Polygons.tsx
+++ b/src/admin/components/ResourceTabs/PolygonReviewTab/components/Polygons.tsx
@@ -15,6 +15,8 @@ import Icon from "@/components/extensive/Icon/Icon";
 import { IconNames } from "@/components/extensive/Icon/Icon";
 import ModalConfirm from "@/components/extensive/Modal/ModalConfirm";
 import { ModalId } from "@/components/extensive/Modal/ModalConst";
+import Pagination from "@/components/extensive/Pagination";
+import { VARIANT_PAGINATION_DASHBOARD } from "@/components/extensive/Pagination/PaginationVariant";
 import { useBoundingBox } from "@/connections/BoundingBox";
 import { useMapAreaContext } from "@/context/mapArea.provider";
 import { useModalContext } from "@/context/modal.provider";
@@ -69,6 +71,9 @@ const Polygons = (props: IPolygonProps) => {
   const [openCollapseAll, setOpenCollapseAll] = useState(false);
   const [currentPolygonUuid, setCurrentPolygonUuid] = useState<string | undefined>(undefined);
 
+  const [currentPage, setCurrentPage] = useState(1);
+  const [pageSize, setPageSize] = useState(20);
+
   const containerRef = useRef<HTMLDivElement>(null);
   const { polygonFromMap, setPolygonFromMap, mapFunctions, siteUuid, isLoading = false } = props;
   const { map } = mapFunctions;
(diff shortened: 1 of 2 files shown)


src/admin/components/ResourceTabs/PolygonReviewTab/components/Polygons.tsx
```diff
@@ -92,6 +95,38 @@ const Polygons = (props: IPolygonProps) => {
...
+      }
+
+      return status === validFilter;
+    });
+  }, [polygonMenu, validFilter]);
+
+  const totalPages = Math.max(1, Math.ceil(filteredPolygons.length / pageSize));
+  const startIndex = (currentPage - 1) * pageSize;
+  const endIndex = startIndex + pageSize;
+  const currentPagePolygons = filteredPolygons.slice(startIndex, endIndex);
```
roguenet, line 118: This should probably get wrapped in the same useMemo that generates `filteredPolygons`, having that one memo return both.

src/admin/components/ResourceTabs/PolygonReviewTab/components/Polygons.tsx
```diff
@@ -69,6 +69,9 @@ const Polygons = (props: IPolygonProps) => {
   const [openCollapseAll, setOpenCollapseAll] = useState(false);
   const [currentPolygonUuid, setCurrentPolygonUuid] = useState<string | undefined>(undefined);
 
+  const [currentPage, setCurrentPage] = useState(1);
+  const pageSize = 20;
```
roguenet, line 73: Please define this as a const outside of the component:
```
const PAGE_SIZE = 20;
```

src/admin/components/ResourceTabs/PolygonReviewTab/components/Polygons.tsx
```diff
@@ -92,6 +95,38 @@ const Polygons = (props: IPolygonProps) => {
...
+  const filteredPolygons = useMemo(() => {
+    if (!validFilter || validFilter === "all") {
+      return polygonMenu;
+    }
+
+    return polygonMenu.filter(polygon => {
+      const status = polygon.validationStatus;
+
+      if (validFilter === "not_checked") {
+        return !status || status === "" || status === "undefined" || status === "null" || status === "notChecked";
```
roguenet, line 108: Please either use `status == null` instead of `!status` or condense this to cover both null and empty string with `isEmpty(status)` from lodash. 

When checking against several string values I sometimes use array functions instead. So another approach would be to define an array outside the component:
```
const INVALID_STATUSES = ["undefined", "null", "notChecked"];
```
and then this line becomes
```
return isEmpty(status) || INVALID_STATUSES.includes(status)
```

src/admin/components/ResourceTabs/PolygonReviewTab/components/Polygons.tsx
```diff
@@ -92,6 +95,38 @@ const Polygons = (props: IPolygonProps) => {
   } = contextMapArea;
 
   const polygonMenu = useMemo(() => props.menu || [], [props.menu]);
+
+  const filteredPolygons = useMemo(() => {
+    if (!validFilter || validFilter === "all") {
```
roguenet, line 100: `validFilter == null || validFilter === "all"`

src/components/elements/MapSidePanel/MapSidePanel.tsx
```diff
@@ -55,9 +55,35 @@ const MapSidePanel = ({
...
+
+  useEffect(() => {
+    if (currentPage > totalPages && totalPages > 0) {
+      setCurrentPage(totalPages);
+    }
+  }, [currentPage, totalPages]);
+
+  useEffect(() => {
+    setCurrentPage(1);
+  }, [checkedValues]);
```
roguenet, line 85: I'm seeing a lot of duplicated code in this component from the one above. Please create a hook to manage this pagination approach and use in both places.

src/admin/components/ResourceTabs/PolygonReviewTab/components/Polygons.tsx
cesarLima1: Done.

src/admin/components/ResourceTabs/PolygonReviewTab/components/Polygons.tsx
cesarLima1: Done.

src/admin/components/ResourceTabs/PolygonReviewTab/components/Polygons.tsx
```diff
@@ -93,6 +97,31 @@ const Polygons = (props: IPolygonProps) => {
...
+  const {
+    currentPage,
+    setCurrentPage,
+    totalPages,
+    startIndex,
+    endIndex,
+    currentPageItems: currentPagePolygons
+  } = usePolygonsPagination(filteredPolygons, [validFilter]);
+
+  // pagination handled by usePagination
```
Scriptmatico, line 123: Just as a comment, I don't think this comment is necessary.

src/hooks/usePolygonsPagination.ts
```diff
@@ -0,0 +1,43 @@
...
+  setCurrentPage: React.Dispatch<React.SetStateAction<number>>;
+  totalPages: number;
+  startIndex: number;
+  endIndex: number;
+  currentPageItems: T[];
+}
+
+export function usePolygonsPagination<T>(
+  items: T[],
+  pageSize: number,
```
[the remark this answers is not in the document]
cesarLima1: Since the page size doesn't need to be variable at the moment, and no changes are expected in the short term. Therefore, I’ve set the pageSize to 20 within the hook.

Scriptmatico (dismissed):

PR #1387, opened 2025-08-05
[TM-2359] add label validation for planting status for admin

cesarLima1:
https://gfw.atlassian.net/browse/TM-2359


src/components/elements/MapPolygonPanel/MapMenuPanelItem.tsx
```diff
@@ -39,6 +39,7 @@ export interface MapMenuPanelItemProps extends DetailedHTMLProps<HTMLAttributes<
   primary_uuid?: string;
   isCollapsed?: boolean;
   validationStatus?: string;
+  isAdmin?: boolean;
```
roguenet, line 42: No need to open a new PR to change this, but for future reference: it's usually cleaner to use the hook in each component than to increase the props that are being passed around.

PR #1394, opened 2025-08-08
[TM-1274] Small updates to workday table UI

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1274


src/components/extensive/DemographicsCollapseGrid/DemographicsRow.tsx
```diff
@@ -34,6 +34,17 @@ const DemographicsRow = ({
 }: DemographicsRowProps) => {
   const [focused, setFocused] = useState(false);
   const t = useT();
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const handleClick = useCallback(() => {
+    if (inputRef.current) {
```
roguenet, line 40: Please use explicit null checks `inputRef.current != null`

PR #1439, opened 2025-09-02
[TM-2439] add condition for replanting project reports

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-2439)

### DIFF
--- a/src/components/extensive/Tables/TreeSpeciesTable/hooks.ts
+++ b/src/components/extensive/Tables/TreeSpeciesTable/hooks.ts
@@ -4,7 +4,7 @@ import { useMemo } from "react";
 
 import { getTreeSpeciesColumns, TableType } from "@/components/extensive/Tables/TreeSpeciesTable/columnDefinitions";
 import { PlantData } from "@/components/extensive/Tables/TreeSpeciesTable/index";
-import { SupportedEntity, usePlants } from "@/connections/EntityAssociation";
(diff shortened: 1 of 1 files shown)


src/components/extensive/Tables/TreeSpeciesTable/hooks.ts
```diff
@@ -70,12 +70,17 @@ export const usePlantTotalCount = ({ entity, entityUuid, collection }: Aggregate
...
+      return sumBy(Object.values(reportCounts ?? {}), "amount");
+    }
+
     if (entity.endsWith("Reports")) {
       return sumBy(plants, "amount");
     } else {
       return sumBy(Object.values(reportCounts ?? {}), "amount");
     }
-  }, [entity, plants, reportCounts]);
+  }, [entity, plants, reportCounts, collection]);
```
roguenet, line 83: :art: This is fine as-is, but you could consider restructuring a little to avoid duplicate code:
```
const usesReportCounts = !entity.endsWith("Reports") ||
  // Special case: for projectReports with "replanting" collection, data comes within reportCounts
  entity === "projectReports" && collection === "replanting";
const records = usesReportCounts ? Object.value(reportCounts ?? {}) : plants;
return sumBy(records, "amount");
```

LimberHope: updated

roguenet (approved):

PR #1467, opened 2025-09-18
[TM-2337] add validation connection

egrojMonroy:


src/connections/Validation.ts
```diff
@@ -0,0 +1,142 @@
...
+const hasValidParams = (pathParams: GetPolygonValidationPathParams | undefined): boolean => {
+  const isValid =
+    pathParams != null &&
+    pathParams.polygonUuid != null &&
+    pathParams.polygonUuid !== "" &&
+    pathParams.polygonUuid !== "undefined" &&
+    !isEmpty(pathParams.polygonUuid);
+
+  return isValid;
+};
```
roguenet, line 24: There's no reason to assign the value and then do nothing but return it (my editor actually warns about this), so this can be a simple inline function:
```
const hasValidParams = (pathParams: GetPolygonValidationPathParams | undefined): boolean =>
  pathParams != null &&
  pathParams.polygonUuid != null &&
  pathParams.polygonUuid !== "" &&
  pathParams.polygonUuid !== "undefined" &&
  !isEmpty(pathParams.polygonUuid);
```

roguenet: This can also be a lot simpler. The isEmpty term catches a lot of whats above, so this might more simply be:
```
!isEmpty(pathParams?.polygonUuid) && pathParams?.polygonUuid !== "undefined"
```

src/connections/Validation.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export const usePolygonValidation = (pathParams: GetPolygonValidationPathParams) => {
+  const result = useConnection(validationConnection, {
+    id: pathParams.polygonUuid,
+    enabled: hasValidParams(pathParams)
+  });
+
+  const validationData = result[1].data;
+
+  return validationData;
```
roguenet, line 42: Please simply `return result[1].data;`

roguenet: Even more straightfoward would be:
```
const [, { data }] = useConnection(...);
return data;
```
or 
```
return useConnection(...)[1].data;
```

src/connections/Validation.ts
```diff
@@ -0,0 +1,142 @@
...
+  .index<ValidationDto>(() => ({
+    pathParams: { siteUuid: "" },
+    queryParams: { page: {} }
+  }))
+  .pagination()
+  .enabledProp()
+  .addProps<{ siteUuid: string; criteriaId?: number }>(({ siteUuid, criteriaId }) => ({
+    pathParams: { siteUuid },
+    queryParams: { page: {}, ...(criteriaId != null && { criteriaId }) }
+  }))
```
roguenet, line 55: This is OK, but it's a little more in keeping with other factory uses to define this in the call to `.index`:
```
.index<ValidationDto, { siteUuid: string, criteriaId?: number }>(({ siteUuid, criteriaId }) => ({ 
  pathParams: { siteUuid },
  queryParams: { page: {}, criteriaId }
})
```

roguenet: It's a little weird to always pass an empty object for page. Is that correct? I think `pagination()` factory method should be handling that for you, so you can likely just leave it off here.

src/connections/Validation.ts
```diff
@@ -0,0 +1,142 @@
...
+
+const ALL_VALIDATIONS_PAGE_SIZE = 100;
+
+export const useAllSiteValidations = (siteUuid: string, criteriaId?: number) => {
+  const [allValidations, setAllValidations] = useState<ValidationDto[]>([]);
+  const [total, setTotal] = useState(0);
+
+  const fetchAllValidationPages = useCallback(
+    async (clearCache: boolean = false) => {
+      if (!siteUuid) return;
```
roguenet, line 66: `== null`

src/connections/Validation.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        const firstPageResponse = await loadConnection(siteValidationConnection, {
+          siteUuid,
+          criteriaId,
+          pageSize: ALL_VALIDATIONS_PAGE_SIZE,
+          pageNumber: 1,
+          enabled: true
+        });
+
+        if (firstPageResponse.loadFailure) {
```
roguenet, line 90: `!= null`

src/connections/Validation.ts
egrojMonroy: done

roguenet (dismissed):

PR #1486, opened 2025-09-27
[TM-2414] v3 Get form

roguenet:
https://gfw.atlassian.net/browse/TM-2414


src/admin/components/Dialogs/FormQuestionPreviewDialog.tsx
```diff
@@ -1,97 +1,76 @@
...
-export const preparePreviewField = (_question: FormQuestionRead, linkedFieldData: FormQuestionField[]) => {
-  const question = questionPayloadToFormQuestionRead(
-    normalizeQuestionCreatePayload(_question, linkedFieldData),
-    linkedFieldData
-  );
-
-  question.uuid = question.uuid || "preview-" + Math.floor(Math.random() * 100000);
-  return { ...question, ...question.additional_props };
+  return null;
+  // const formHook = useForm();
```
roguenet, line 20: There are a few places like this with commented out code. That will all be fixed up before this merges to staging, and in most cases I only did this in places where I know I'm going to address it in the next ticket.

src/pages/entity/[entityName]/create/[frameworkKey].page.tsx
```diff
@@ -78,34 +65,34 @@ const EntityIntroPage = () => {
...
+  }, [createEntity, entityName, entityUUID, formUUID, parentName, parentUUID, router]);
 
   return (
     <BackgroundLayout>
       <ContentLayout>
         <LoadingContainer loading={!form}>
-          <WizardFormIntro
-            title={form?.title!}
-            //@ts-ignore
-            imageSrc={form?.banner?.url}
```
Comment (roguenet) [inline src/pages/entity/[entityName]/create/[frameworkKey].page.tsx:90]: I discovered this morning that this was being added via `appendFilesToResources` in PHP (I had thought it was simply missing, and the `ts-ignore` was hiding that fact). I'm going to make sure this gets added back in in my current ticket (TM-2415 - indexForms)

Comment (roguenet) [reply src/pages/entity/[entityName]/create/[frameworkKey].page.tsx:90]: Addressed here: https://github.com/wri/wri-terramatch-website/commit/deaf3e987c376e6797781712d253c2a3980e5d50

PR #2180, opened 2026-04-07
[TM-3217] Refactor banner logo functionality

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3217)


src/redesignComponents/content/headers/PageHeaders/components/TeamSection.tsx
```diff
@@ -22,6 +23,7 @@ export interface TeamSectionProps {
 
 const TeamSection: FC<TeamSectionProps> = ({ team, onAddTeamClick, gotoTeamMembers, frameworkKey }) => {
   const t = useT();
+  const bannerLogoKind = getProjectBannerLogoKind(frameworkKey);
```
roguenet: This kind of props drilling is exactly what the framework provider was designed to avoid. It's frequently true that a given page will have components all over the place that need to know what the current framework is. This should be implemented like this instead:

1. `<FrameworkProvider>` should be used at the top level project page, wrapping the whole render tree as soon as the project has been retrieved from the API.
2. These logo components should either use the `withFrameworkShow` HOC (see example usage in the codebase) or their usage should be surrounded `<ContextCondition>`
3. Then the lines below become one of these depending on your choice for item 2:
```
<TerraFundAFR100 frameworksShow={TERRAFUND_AFR100_BANNER_LOGO_FRAMEWORKS} className="mt-auto" />
<HaritBharatFund frameworksShow={[Framework.HBF]} className="mt-auto" />
```
or 
```
<ContextCondition frameworksShow={TERRAFUND_AFR100_BANNER_LOGO_FRAMEWORKS}>
  <TerraFundAFR100 fclassName="mt-auto" />
</ContextCondition>
<ContextCondition frameworksShow={[Framework.HBF]}>
  <HaritBharatFund  className="mt-auto" />
</ContextCondition>
```
4. Remove `frameworkKey` from the props of this component, and probably all of its parents as well. All project page components that need to know the framework should get it from the context instead of a prop.

roguenet (commented):
I think this is still missing some pieces:
* I was expecting `project/[uuid]/index.page.tsx` to add the framework provider
* I was expect `TeamSection.tsx` to remove its dependence on the framework key prop (and any other parents up the tree to the project page)

src/redesignComponents/content/headers/PageHeaders/components/TeamSection.tsx
Marcolr-dev: Thank you for the detailed guide; I have adapted the implementation to that pattern.
FrameworkProvider 
`src/pages/project/[uuid]/index.page.tsx` wraps the project UI in `<FrameworkProvider frameworkKey={project.frameworkKey}>` as soon as `useFullProject` returns a project (same tree as `ProjectContent`, loader overlay, etc.).

No `frameworkKey` prop drilling
`TeamSection` no longer accepts `frameworkKey`. Parents up to the project page (`ProjectBanner` → `ProjectHeader` → `TeamSection`) do not pass framework; anything that needs the current framework reads it via context (`useFrameworkContext` / `ContextCondition`).

Logos: `ContextCondition` (same mechanism as `withFrameworkShow`)
We used the `<ContextCondition frameworksShow={…}>` wrapper, which is implemented as `withFrameworkShow` in `src/context/ContextCondition.tsx` — so this matches your second option (equivalent to wrapping each logo with the HOC), without putting `frameworksShow` on the logo components themselves. TerraFund AFR100, HBF, and PPC each sit behind their own `ContextCondition` with the appropriate `frameworksShow` list (`TERRAFUND_AFR100_BANNER_LOGO_FRAMEWORKS`, `[Framework.HBF]`, `[Framework.PPC]`).

If you’d strongly prefer the logos themselves to be created with `withFrameworkShow(TerraFundAFR100)` etc., we can refactor to that style in a follow-up; behavior would stay the same.

Marcolr-dev:
> I think this is still missing some pieces:
> 
> * I was expecting `project/[uuid]/index.page.tsx` to add the framework provider
> * I was expect `TeamSection.tsx` to remove its dependence on the framework key prop (and any other parents up the tree to the project page)

Those pieces should be in place on the latest revision:

1. **`project/[uuid]/index.page.tsx`** — After the project is loaded from the API, the render tree is wrapped with `<FrameworkProvider frameworkKey={project.frameworkKey}>` (see `ProjectDetailPage` in that file).

2. **`TeamSection.tsx`** — The component no longer takes a `frameworkKey` prop; its props are only team-related handlers and data. Programme logos are gated with `ContextCondition` + `frameworksShow`, reading the active framework from `FrameworkProvider`.

3. **Parents** — `ProjectHeader` renders `<TeamSection … />` without a framework prop; nothing between the project page and `TeamSection` drills `frameworkKey` for the banner logos.

If the PR diff still shows the older `frameworkKey` prop or manual logo branching, try refreshing the review or comparing against the latest commit on this branch — happy to walk through the exact lines if anything still looks off on your side.

roguenet (approved):

PR #2188, opened 2026-04-08
[TM-3145] Refactor financial metrics and funding sources components t…

Marcolr-dev:

### DIFF
--- a/src/components/extensive/WizardForm/fields/number.field.tsx
+++ b/src/components/extensive/WizardForm/fields/number.field.tsx
@@ -1,6 +1,6 @@
 import cn from "classnames";
 import { isNumber } from "lodash";
-import { useCallback, useEffect, useId, useState } from "react";
+import { type FC, useCallback, useEffect, useId, useState } from "react";
 import { useController } from "react-hook-form";
 import * as yup from "yup";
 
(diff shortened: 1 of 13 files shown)


src/components/extensive/WizardForm/fields/number.field.tsx
```diff
@@ -1,46 +1,190 @@
...
+  }
+  const n = typeof value === "number" ? value : Number(value);
+  if (!Number.isFinite(n)) {
+    return "";
+  }
+  return formatFinancialAmount(n, currencyCode);
+}
+
+/** Locale-aware amount input for wizard number fields; keeps RHF value as a number for the API. */
+function FinancialAmountInput(props: SharedFieldProps) {
```
roguenet: use `const FinancialAmountInput: FC<ShareFieldProps> = props => {` please.

src/components/extensive/WizardForm/fields/number.field.tsx
```diff
@@ -1,46 +1,190 @@
...
+      label={label}
+      description={description}
+      error={error}
+      required={required}
+      feedbackRequired={feedbackRequired}
+    >
+      <input
+        id={id}
+        type="text"
+        inputMode="decimal"
```
roguenet, line 90: Ahh, is this the magic that makes the local browser input use `,` instead of `.` for decimals when appropriate?

roguenet (commented):
One tiny change for consistency, but looks good overall. 👍🏻

src/components/extensive/WizardForm/fields/number.field.tsx
Marcolr-dev: Not quite. inputMode="decimal" doesn’t make the browser automatically swap commas for dots (or the other way around); it mainly nudges mobile browsers to show a numeric keypad that’s suited to decimal entry. Because we’re using type="text", the visible formatting (comma vs dot, grouping, etc.) is entirely driven by our controlled text state plus formatFinancialAmount / parseFinancialAmountInput for the currency/locale. The value we send to the API stays numeric through RHF once we commit on blur.

roguenet (approved):

PR #2211, opened 2026-04-13
[TM-3145] Enhance InformationTab and Details components with orgDetai…

Marcolr-dev:


src/admin/components/ResourceTabs/InformationTab/index.tsx
```diff
@@ -104,6 +108,8 @@ const InformationTab: FC<IProps> = props => {
     [fieldsProvider]
   );
 
+  const [orgDetailsLoaded, orgDetails] = useV2OrgFormDetails(record?.organisationUuid ?? undefined);
```
roguenet: The name of this hook is out of date. Please change it to simply `useOrgFormDetails` (the base hook name and all its uses)

src/components/extensive/WizardForm/types.ts
```diff
@@ -17,6 +17,8 @@ import { CSVGenerator } from "@/utils/CsvGeneratorClass";
 
 export type FieldInputType =
   | FormQuestionDto["inputType"]
+  // Monetary amount fields (TM-2925); keep explicit so Record<FieldInputType, …> stays valid if codegen lags OpenAPI
+  | "number-currency"
```
roguenet: I've mentioned this before, but please don't leave in code that is just trying to get around codegen lag. It's correct for a FE build to fail until the BE branches have been deployed that it depends on. Putting code like this in place could end up with us sending a branch to the FE that doesn't work because it's depending on BE code that isn't in place yet.

roguenet (approved):

PR #2218, opened 2026-04-14
[TM-3269] remove logout error and list only projects in project card

egrojMonroy:
https://gfw.atlassian.net/browse/TM-3269


cesarLima1 (dismissed):

roguenet (commented):
This seems almost as confusing to the user as the logout. Definitely an improvement, but maybe a toast should be added as well?

roguenet (approved):

PR #2220, opened 2026-04-15
[TM-3145] Add support for project budget integer fields in NumberFiel…

Marcolr-dev:

### DIFF
--- a/src/components/extensive/WizardForm/fields/number.field.tsx
+++ b/src/components/extensive/WizardForm/fields/number.field.tsx
@@ -12,6 +12,7 @@ import { useCurrencyContext } from "@/context/currency.provider";
 import { useWizardOrgFormDetails } from "@/context/wizardForm.provider";
 import {
   formatFinancialAmount,
+  isProjectBudgetIntegerField,
   parseFinancialAmountInput,
   shouldFormatFinancialNumberField
 } from "@/utils/financialReport";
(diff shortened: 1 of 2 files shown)


pachonjcl (approved):

src/components/extensive/WizardForm/fields/number.field.tsx
```diff
@@ -154,6 +165,9 @@ export const NumberField: FormFieldFactory = {
     if (shouldFormatFinancialNumberField(field)) {
       return <FinancialAmountInput {...sharedProps} />;
     }
+    if (isProjectBudgetIntegerField(field)) {
+      return <Input {...sharedProps} type="number" step={1} />;
```
roguenet, line 169: Why is this skipping additional props? Instead, the additional props should get correctly assigned in the form editor. I'm going to approve this because we need to get this release out the door, but all this customization based on the specific linked field is not how this system is designed to work. The FE should really _never_ be customizing something based on linked field key and it's been happening quite a lot. 

There are a few other cases I need to address (like financial indicators added a whole new column to the form question table instead of using additional props). All customization of form questions should be handled in additional props and the FE should be able to render correctly based entirely on the input type, additional props, collection, etc that come from the DB. Doing it based on linked field key is fragile and prone to hidden bugs sneaking out.

roguenet (approved):

PR #2234, opened 2026-04-17
[TM-3283] financial report fields not showing up on change requests

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-3283)

### DIFF
--- a/src/components/extensive/WizardForm/FormSummaryRow/getFormEntries.tsx
+++ b/src/components/extensive/WizardForm/FormSummaryRow/getFormEntries.tsx
@@ -107,7 +107,8 @@ export const getFormEntries = (
       entries.push({
         title: field.label ?? "",
         inputType: field.inputType,
-        value: getFormattedAnswer(field, values, fieldsProvider) ?? nullText ?? t("Answer Not Provided")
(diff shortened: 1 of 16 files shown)


src/components/extensive/WizardForm/FormSummaryRow/getFormEntries.tsx
```diff
@@ -107,7 +107,8 @@ export const getFormEntries = (
       entries.push({
         title: field.label ?? "",
         inputType: field.inputType,
-        value: getFormattedAnswer(field, values, fieldsProvider) ?? nullText ?? t("Answer Not Provided")
+        value: getFormattedAnswer(field, values, fieldsProvider) ?? nullText ?? t("Answer Not Provided"),
+        optionsList: field.optionsList ?? null
```
roguenet: I think this general approach is not great. `getFormEntries` should be able to provide the final value without requiring additional processing in e.g. `ChangeRow` - I wouldn't expect to see any changes to that component for this fix. See slack for further discussion - I want to understand what you're trying to accomplish before I can suggest a better way.

pachonjcl (approved):

PR #2262, opened 2026-04-22
[TM-3288] Enhance user data provider and user show components with pr…

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3288)


LimberHope (dismissed):

src/admin/apiProvider/dataProviders/userDataProvider.ts
```diff
@@ -95,12 +97,24 @@ export const userDataProvider: Partial<DataProvider> = {
   },
 
   async getOne<RecordType>(_: string, { id }: GetOneParams) {
+    const uuid = id as string;
+    const meUserId = ApiSlice.currentState.meta.meUserId;
+    // List/index cache can satisfy `loadUser` with a row missing detail-only attributes (e.g.
+    // monitoringPartnerProjects). Drop the cached user so the detail GET runs for Show — except
+    // for the logged-in user: that entry is `users[meUserId]` and is required by `useMyUser` /
+    // `selectMe`. Pruning it removes the session user from the store, unmounts `<Admin>` (no
+    // resources), and floods the menu with "Resource not found" (super admins often open their own profile).
```
roguenet: This should be handled with the `lightResource` pattern we use for entities. Let me know if you have any questions about implementing it.

roguenet: See my comment on the [BE PR](https://github.com/wri/terramatch-microservices/pull/701#discussion_r3132295967), I think simply including the projects on all users is better than the `lightResource` pattern in this case.

src/admin/apiProvider/dataProviders/userDataProvider.ts
```diff
@@ -95,12 +96,19 @@ export const userDataProvider: Partial<DataProvider> = {
   },
 
   async getOne<RecordType>(_: string, { id }: GetOneParams) {
+    const uuid = id as string;
+    const meUserId = ApiSlice.currentState.meta.meUserId;
+    if (meUserId == null || uuid !== meUserId) {
+      ApiSlice.pruneCache("users", [uuid]);
+    }
```
roguenet: Why is this needed?

Marcolr-dev: The user connection layer can serve a light cached user for loadUser without refetching a full detail payload
For the admin User Show we need the full record (including fields like monitoringPartnerProjects that may be missing on the cached copy), so we prune that id from the store before loadUser to force a fresh fetch.

We skip pruning when uuid === meUserId because that cache entry backs useMyUser / the logged-in session. Clearing it caused the app to treat the current user as unloaded and broke the admin shell (e.g. white screen / “resource not found” when opening your own profile as super admin)

roguenet: This is why I had you include the new properties in the index view in the BE. The object being returned from the BE should be the same for the get and index views unless I'm missing something?

Marcolr-dev: I'll do the test right now

Marcolr-dev: @roguenet Agreed, with the index returning the same attributes as GET, the prune was redundant. Removed it and the meUserId exception; getOne now just uses loadUser.

roguenet (approved):

PR #2591, opened 2026-06-15
[TM-3554] Project Report Overview (design+dev)

cesarLima1:
https://gfw.atlassian.net/browse/TM-3554


src/pages/reports/project-report/components/KeyIndicatorsInsights.tsx
```diff
@@ -0,0 +1,193 @@
...
+        frameworkKey={framework}
+      />
+    </>
+  );
+
+  const renderCards = () => {
+    if (framework === Framework.PPC) return renderPpcCards();
+    if (framework === Framework.HBF) return renderHbfCards();
+    if (TERRAFUND_FRAMEWORKS.includes(framework)) return renderTerrafundCards();
+    return renderTerrafundCards();
```
roguenet: You're rendering TF in both of these cases, so the `TERRAFUND_FRAMEWORKS.includes` clause is not doing anything. More than that though, this method of choosing what to render is a bit of a code smell. It's going to have performance problems because the results of `renderCards()` below is freshly calculated on ever render of this component. A better approach would be to make the content of `renderPpcCards`, `renderHbfCards` and `renderTerrafundCards` into components, and selectively render the correct component below.

egrojMonroy: Separated into components, and removed redundancy.

pachonjcl (dismissed):

LimberHope (approved):

PR #2700, opened 2026-06-26
[TM-3629] Fix assumption that props.models is always a single model.

roguenet:
https://gfw.atlassian.net/browse/TM-3629


src/components/extensive/WizardForm/index.tsx
```diff
@@ -611,7 +633,7 @@ function WizardForm(props: WizardFormProps) {
                   <ResponsiveBreadcrumbToolbar
                     breadcrumbs={linkHeaderMap}
                     suffix={
-                      formModel.model.includes("Reports") && (
+                      models[0]?.model.includes("Reports") && (
```
roguenet, line 636: This was the specific spot that caused the bug, but I went ahead and cleared up the bad assumption about `props.models` everywhere.

cesarLima1 (approved):

